PR #19, opened 2021-08-31
review

SergioDemianLerner:


contracts/LiquidityBridgeContract.sol
```diff
@@ -409,11 +517,17 @@ contract LiquidityBridgeContract {
             s := mload(add(signature, 0x40))
             v := byte(0, mload(add(signature, 0x60)))
         }
+		// SDL: I'm think we should not use this signature format
+		// but an EIP712 compatible format.
```
javicid, line 521: This is compatible with web3. I would have to read EIP712 and see if it can be easily implemented in Go/web3. I suggest to leave this as future work.

contracts/LiquidityBridgeContract.sol
```diff
@@ -409,11 +517,17 @@ contract LiquidityBridgeContract {
...
         bytes memory prefix = "\x19Ethereum Signed Message:\n32";
         bytes32 prefixedHash = keccak256(abi.encodePacked(prefix, hash));
         return ecrecover(prefixedHash, v, r, s) == addr;
     }
 
+	// SDL: Sadly this is not the format that I designed for the bridge.
+	// I designed it to be a double hash but it's a single hash.
+	// We may have problems in the future introducing versioning, because
+	// of the variable size.
```
javicid, line 530: Can this be solved by hashing the quote again?

contracts/LiquidityBridgeContract.sol
```diff
@@ -10,30 +10,34 @@ import './Bridge.sol';
...
         bytes20 fedBtcAddress;
         address lbcAddress;
         address liquidityProviderRskAddress;
         bytes btcRefundAddress;
         address rskRefundAddress;
         bytes liquidityProviderBtcAddress;
-        uint callFee;
-        uint penaltyFee;
+		// SDL: Versioning field is missing. Because no double-hash was used,
+		// then now the version field would need to go here I think
```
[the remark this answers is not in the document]
SergioDemianLerner: By mistake, we didn't add a version field to the flyover protocol in the Bridge (Iris HF). Now we must find the best place to store it. The problem is that now, because the native Bridge arguments are a prefix of the full msg, we cannot distinguish old from new msgs by size. We can talk about this in a call

contracts/LiquidityBridgeContract.sol
```diff
@@ -10,30 +10,34 @@ import './Bridge.sol';
...
+        uint callFee; // SDL: Can't we re-scale to reduce field size?
+        uint penaltyFee; // SDL: Can't we re-scale to reduce field size?
```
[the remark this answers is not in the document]
SergioDemianLerner: uint is a very very long number (32 bytes). Every byte you add to the calldata costs bitcoins. In the future that will be expensive. It's a luxury to waste bytes.

contracts/LiquidityBridgeContract.sol
```diff
@@ -10,30 +10,34 @@ import './Bridge.sol';
...
+        uint nonce; // SDL: Why not uint32 ?
         uint value;
-        uint agreementTimestamp;
-        uint timeForDeposit;
-        uint callTime;
-        uint depositConfirmations;
+        uint agreementTimestamp; // SDL: Do we really need 2^256 range?
+        uint timeForDeposit; // SDL: Do we really need 2^256 range?
+        uint callTime; // SDL: Do we really need 2^256 range?
+        uint depositConfirmations; // SDL: Do we really need 2^256 range?
```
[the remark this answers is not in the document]
SergioDemianLerner: good question. Bitcoin timestamps are 32 bits. (Will overflow in 2106). With 40 bits we have enough before the world self-destructs.  
But don't cast silently. Make sure that the cast reverts if value is out of range

contracts/LiquidityBridgeContract.sol
```diff
@@ -9,31 +9,52 @@ import './Bridge.sol';
...
     }
 
     struct Registry {
-        uint256 timestamp;  
+        uint32 timestamp;  // SDL: if we reduce range to uint32 then both fields fit into a single 256-bit cell, reducing gas cost 50%
```
SergioDemianLerner, line 53: I can also be uint40, if you opt to use 40-bit timestamps.

contracts/LiquidityBridgeContract.sol
```diff
@@ -193,15 +197,30 @@ contract LiquidityBridgeContract {
         @param quote The quote that identifies the service
         @return Boolean indicating whether the call was successful
      */
+	 // SDL I don't like this method to be able to be called recursively
+	 // even if I can't see a vuln in the  logic. Defensive programming should
+	 // be applied.
```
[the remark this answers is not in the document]
SergioDemianLerner: I would add a function modifier enterOnlyOnce() that increments a state variable, and reverts if the state variable was not zero on entry. And I think you can use also a function modifier callWhenExit() that decrements the value.

contracts/LiquidityBridgeContract.sol
```diff
@@ -193,15 +197,30 @@ contract LiquidityBridgeContract {
...
+		// should be performed in ints instead of uints (unless Solidity is
+		// already casting the subtraction to int, but I don't think so)
+		// Is there any test case for this ?
         balances[quote.liquidityProviderRskAddress] += msg.value - quote.value;
-        emit BalanceIncrease(quote.liquidityProviderRskAddress, msg.value);
+        
+		// SDL: Is this event really needed? What for ?
+		emit BalanceIncrease(quote.liquidityProviderRskAddress, msg.value);
 
+		// SDL: Shoudn't here by a "gap" added to gasLimit ?
```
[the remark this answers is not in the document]
SergioDemianLerner: I mean, shouldn't this be something like:

require(gasleft() +someExtraGas>= quote.gasLimit, "Insufficient gas");

contracts/LiquidityBridgeContract.sol
```diff
@@ -215,6 +256,7 @@ contract LiquidityBridgeContract {
             balances[quote.liquidityProviderRskAddress] += quote.value;
         }
         emit CallForUser(msg.sender, quote.contractAddress, quote.gasLimit, quote.value, quote.data, success, quoteHash);
+        processedQuotes[quoteHash] = CALL_DONE_CODE;
```
SergioDemianLerner, line 259: I must review this new processedQuotes thing

contracts/LiquidityBridgeContract.sol
```diff
@@ -227,6 +246,11 @@ contract LiquidityBridgeContract {
         @param height The block that contains the peg-in transaction
         @return The total peg-in amount received from the bridge contract or an error code
      */
+	 
+	 //SDL: No reentrancy guard. I see that registerBridge() does not allow
+	 // reentrancy, but that not an explicit guard. Maybe it's better to 
+	 // ad a guard as defensive programming.
+
```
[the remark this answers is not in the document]
SergioDemianLerner: responded in a preview thread

contracts/LiquidityBridgeContract.sol
```diff
@@ -235,35 +259,66 @@ contract LiquidityBridgeContract {
         uint256 height
     ) public returns (int256) {
         bytes32 quoteHash = hashQuote(quote);
+		
+		// SDL: I don't remember why this should be signed...
```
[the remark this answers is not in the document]
SergioDemianLerner: yes!

contracts/LiquidityBridgeContract.sol
```diff
@@ -235,35 +259,66 @@ contract LiquidityBridgeContract {
...
         require(transferredAmount != -304, "Invalid transaction value");
 
+		
         if (shouldPenalize(quote, transferredAmount, callRegistry[quoteHash].timestamp, height)) {
             collateral[quote.liquidityProviderRskAddress] -= quote.penaltyFee;
             emit Penalized(quote.liquidityProviderRskAddress, quote.penaltyFee, quoteHash);
             
             // pay reward to sender
+			// SDL: rewardR is a divisor... why not making more granual, such as 
+			// multiplying penaltyFee by 10 before dividing ?
```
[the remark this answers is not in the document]
SergioDemianLerner: that's what I was saying (although not so clearly) , but generally it's better to multiply to prevent underflows (depends on args size not to overflow, of course)

penaltyFee  * rewardR /100

contracts/LiquidityBridgeContract.sol
SergioDemianLerner: ok. But note that most web3 providers also support EIP712 signatures.

contracts/LiquidityBridgeContract.sol
SergioDemianLerner: Let's discuss this in a call.

contracts/LiquidityBridgeContract.sol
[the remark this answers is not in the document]
javicid: I am not sure, what's the point of the extra gas?

You suggested me to add this check not to ensure that the transaction has enough gas to finish execution, but to guarantee that the LP cannot perform this call with less gas than the agreed amount. This is because RSK will not raise an error when performing a call with less gas left than the defined gas limit.

contracts/LiquidityBridgeContract.sol
```diff
@@ -9,31 +9,53 @@ import './Bridge.sol';
...
         bytes data;        
-        uint gasLimit;
-        uint nonce;
-        uint value;
-        uint agreementTimestamp;
-        uint timeForDeposit;
-        uint callTime;
-        uint depositConfirmations;
+        uint32 gasLimit; // SDL: Do we really need 2^256 range?
+        uint16 nonce; // SDL: Why not uint32 ?
```
SergioDemianLerner, line 45: Sure it will fit in 16 bits ? :)

contracts/LiquidityBridgeContract.sol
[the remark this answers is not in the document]
SergioDemianLerner: I wrote it wrong. I meant

require(gasleft() >= quote.gasLimit +someExtraGas, "Insufficient gas");

contracts/LiquidityBridgeContract.sol
```diff
@@ -225,69 +279,134 @@ contract LiquidityBridgeContract {
...
+		// associated identifier. What's the difference between them?
+		
+		
+        if (transferredAmountOrErrorCode == BRIDGE_REFUNDED_LP_ERROR_CODE || transferredAmountOrErrorCode == BRIDGE_REFUNDED_USER_ERROR_CODE) {
             // Bridge cap exceeded
-            callRegistry[quoteHash].timestamp = 0;
-            callRegistry[quoteHash].success = false;
-            emit BridgeCapExceeded(quoteHash, transferredAmount);
-            return transferredAmount;
+			// There is a Solidity sentence called "delete" that I think
```
SergioDemianLerner, line 344: Remember to check if delete is what you need here

contracts/LiquidityBridgeContract.sol
```diff
@@ -276,13 +331,23 @@ contract LiquidityBridgeContract {
             increaseBalance(quote.liquidityProviderRskAddress, refundAmount);
             int256 remainingAmount = transferredAmount - int(refundAmount);
             
+			// SDL: Should we refund ANY amount. Maybe some wallets do round 
+			// the number to X decimal places and the user cannot send exactly
+			// the amount he wants. If the extra amount sent is "dust", then
+			// maybe we should not attempt to refund.
```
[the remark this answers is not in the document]
SergioDemianLerner: you can set it this way: you take the minimum gas price, and you multiply it by 4500 (50% of the cost of a value transfer call).
This means that overhead of the payment will never be more than 50%.

But it's not a good idea to get the minimum gas price every time from the BlockHeader contract, because getting the minimum has cost of ~4000 gas.  I can't use the current transaction gasPrice, because that is controlled by the caller.
The easier way is just compute DUST once according to the current gasPrice, and leave like that.
Another possibility is to use Bitcoin dust value. (defined in Bitcoin source code)
One really cool but overengineered possibility is to update the DUST variable (According to minimumGasPrice) pseudo-randomly checking some bytes from the previous block hash (using the BLOCKHASH opcode, which costs only 20 gas). 
For example if (prevBlockhash % 100 == 0) recomputeDust()

contracts/LiquidityBridgeContract.sol
SergioDemianLerner: we need to do this call. I think we already have scheduled it

contracts/LiquidityBridgeContract.sol
javicid: My mistake. Replaced to int64 to be consistent with the LP

contracts/LiquidityBridgeContract.sol
javicid: Since I added a check to avoid processing quotes twice this is no longer needed unless there is an advantage in freeing memory.

contracts/LiquidityBridgeContract.sol
```diff
@@ -57,28 +71,42 @@ contract LiquidityBridgeContract {
     mapping(address => uint256) private balances;
     mapping(address => uint256) private collateral;
     mapping(bytes32 => Registry) private callRegistry;  
-    mapping(address => uint256) private resignations;
+    mapping(address => uint256) private resignationBlockNum;
+
+    uint64 private minCollateral;        
+    uint32 private rewardP;     
+    uint32 private resignDelayInBlocks;  
+    int256 private dust;
```
SergioDemianLerner, line 79: Can you add a public getter to inspect the dust value ?

contracts/LiquidityBridgeContract.sol
```diff
@@ -193,28 +221,30 @@ contract LiquidityBridgeContract {
...
         (bool success, ) = quote.contractAddress.call{gas:quote.gasLimit, value: quote.value}(quote.data);
         
-        bytes32 quoteHash = hashQuote(quote);
-        callRegistry[quoteHash].timestamp = block.timestamp;
+        require(block.timestamp <= MAX_UINT32, "Block timestamp overflow");
+        callRegistry[quoteHash].timestamp = uint32(block.timestamp);
 
         if (success) {            
             callRegistry[quoteHash].success = true;
+            balances[quote.liquidityProviderRskAddress] -= quote.value;
```
SergioDemianLerner, line 243: If you have increaseBalance() method, you could also have a balanceDecrease(), and put them one close to the other.

contracts/LiquidityBridgeContract.sol
javicid: done!

contracts/LiquidityBridgeContract.sol
javicid: done!

contracts/LiquidityBridgeContract.sol
```diff
@@ -225,69 +258,86 @@ contract LiquidityBridgeContract {
...
-            emit Refund(quote.rskRefundAddress, transferredAmount, quoteHash);
+        } else {
+            int256 refundAmount = transferredAmountOrErrorCode;
+
+            if (performCall && refundAmount >= int64(quote.value)) {
+                (bool callSuccess, ) = quote.contractAddress.call{gas:quote.gasLimit, value: quote.value}(quote.data);
+                emit CallForUser(msg.sender, quote.contractAddress, quote.gasLimit, quote.value, quote.data, callSuccess, quoteHash);
+
+                if (callSuccess) {
+                    refundAmount -= int64(quote.value);
```
SergioDemianLerner, line 330: Will there be a way to withdraw in the future the sum of all dusts ?
Should we setup a notaryAccount in the contract that is able to withdraw any balance that is stuck on the contract and does not belong to anybody ? I think we should, and also allow this notaryAddress to be changed (only the notary can change it), so we can set it to zero if we see there is no need for it.

SergioDemianLerner: I also think the notaryAddress should be able to move tokens stuck into this contract. It's common for users to mistake addresses and send tokens where they should not. If we don't give a failback method for someone to recover those tokens, they will be lost.

javicid: Ok, but I would leave this for a different PR

contracts/LiquidityBridgeContract.sol
```diff
@@ -225,69 +258,86 @@ contract LiquidityBridgeContract {
...
+     */	 
     function registerPegIn(
         Quote memory quote,
         bytes memory signature,
         bytes memory btcRawTransaction, 
         bytes memory partialMerkleTree, 
-        uint256 height
-    ) public returns (int256) {
+        uint256 height,
+        bool performCall
```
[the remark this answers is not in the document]
javicid: done

contracts/LiquidityBridgeContract.sol
javicid: Another option would be to give this to the LP by increasing its balance

SergioDemianLerner: I'm not sure. We can do that for now. Let's leave this for another PR.

javicid (approved):

PR #21, opened 2021-10-06
Prevents Bridge Contract funds to be stolen

aguschirico:


contracts/LiquidityBridgeContract.sol
```diff
@@ -301,35 +301,38 @@ contract LiquidityBridgeContract {
...
             } else {
-                refundAmount = min(uint(transferredAmountOrErrorCode), quote.callFee);
+                refundAmount = min(transferredAmount, quote.callFee);
             }
             increaseBalance(quote.liquidityProviderRskAddress, refundAmount);
-            int256 remainingAmount = transferredAmountOrErrorCode - int(refundAmount);
+            uint remainingAmount = transferredAmount - refundAmount;
             
-            if (remainingAmount > dust) {
+            if (remainingAmount > uint(dust)) {
```
javicid, line 320: I think that this casting is unnecessary

test/liquidityBridgeContract.js
```diff
@@ -17,15 +17,114 @@ contract('LiquidityBridgeContract', async accounts => {
         let currAddr = web3.eth.currentProvider.getAddress();
         let existing = await instance.getCollateral(currAddr); 
 
-        await instance.register({value : val});
+        await instance.register({value : val, from: web3.eth.currentProvider.addresses[0]});
```
javicid, line 20: Isn't this the default behavior? Also, could use lpRskAddress (I think) variable here instead.

javicid (changes requested):

test/liquidityBridgeContract.js
aguschirico: removed param.

contracts/LiquidityBridgeContract.sol
aguschirico: removed.

test/liquidityBridgeContract.js
```diff
@@ -92,7 +191,7 @@ contract('LiquidityBridgeContract', async accounts => {
 
         assert.equal(currentLPBalance.toNumber(), initialLPBalance.toNumber());
 
-        amount = await instance.registerPegIn.call(
+        amount = await instance.registerPegIn(
```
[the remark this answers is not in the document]
aguschirico: I think this one was a local change I had and for some reason became part of the pr. Changed back.

javicid (approved):

PR #22, opened 2021-10-13
Safely handle underflow and overflow scenarios.

aguschirico:


contracts/LiquidityBridgeContract.sol
```diff
@@ -350,6 +354,10 @@ contract LiquidityBridgeContract {
         return a < b ? a : b;
     }
 
+    function max(uint a, uint b) private pure returns (uint) {
+        return a > b ? a : b;
+    }
+
```
javicid, line 360: I think that solidity already provides a native max function

contracts/LiquidityBridgeContract.sol
```diff
@@ -26,6 +28,7 @@ contract LiquidityBridgeContract {
     int16 constant BRIDGE_UNPROCESSABLE_TX_VALIDATIONS_ERROR = -303;
     int16 constant BRIDGE_UNPROCESSABLE_TX_VALUE_ZERO_ERROR = -304;
     int16 constant BRIDGE_GENERIC_ERROR = -900;
+    uint constant MAX_UINT = 2**256 - 1;
```
javicid, line 31: Is this (2^256) - 1 or 2^255? Shouldn't it be 2^256?

contracts/LiquidityBridgeContract.sol
javicid, line 360: Is this function used?

PR #24, opened 2021-10-19
Tests refactor + ci setup

aguschirico:
image


truffle-config.js
```diff
@@ -1,42 +1,54 @@
...
+        providerOrUrl: `https://public-node.testnet.rsk.co`,
+        derivationPath: "m/44'/37310'/0'/0/"
+      }),
+      network_id: 31,
+      gasPrice: 6000000000,
     },
+    rskMainnet: {
+      provider: () => new HDWalletProvider({
+        mnemonic,
+        providerOrUrl: `https://public-node.rsk.co`,
```
javicid, line 35: Not sure it is a good idea to use a public node here. I was having a lot of connectivity issues when I tried to run the tests against it.

PR #52, opened 2022-10-25
Sync contract with liquidity provider server

joaoiovlabs:


contracts/LiquidityBridgeContract.sol
```diff
@@ -113,7 +112,7 @@ contract LiquidityBridgeContract {
     bool private locked;
 
     mapping(bytes32 => uint8) private processedQuotes;
-    mapping(bytes32 => uint8) private processedPegOutQuotes;
+    mapping(bytes32 => uint8) public processedPegOutQuotes;
```
coolwather, line 115: I think this one should be private like the line above. By security reasons.

joaoiovlabs: when map is public, you can only see the values, you can't change.

coolwather (approved):

MaximStanciu8 (approved):

PR #59, opened 2022-12-07
GBI-1002 - Implement the Upgradeability to LBC

joaoiovlabs:


coolwather (approved):

manfioLP (commented):
I might be wrong, but after this upgrade, we need to add the `delegateCall` functionality, if not, the caller would be the proxy rather than the contract itself
https://blog.openzeppelin.com/the-state-of-smart-contract-upgrades/#upgrade-patterns

PR #71, opened 2022-12-13
Fixed package lock

coolwather:


manfioLP (approved):
LGTM, not much to add since we need package and package-lock in sync

PR #81, opened 2023-02-16
GBI-1112

Dominikkq:


coolwather (approved):

contracts/LiquidityBridgeContract.sol
```diff
@@ -185,7 +185,9 @@ contract LiquidityBridgeContract is Initializable {
     receive() external payable {
         require(msg.sender == address(bridge), "Not allowed");
     }
-
+    function getProviderIds() external view returns (uint) {
+        return providerId;
```
MaximStanciu8, line 189: something weird here,
method says Ids, but return Id

MaximStanciu8 (changes requested):

contracts/LiquidityBridgeContract.sol
Dominikkq: Since provider ids are incremented +1 each time, by returning a single number we can create the array on server-side leaving less code on the contract. If the problem is semantics i can rename.

MaximStanciu8 (approved):

PR #100, opened 2023-04-24
Add new pegout deposit method

Luisfc68:


contracts/LiquidityBridgeContract.sol
```diff
@@ -700,17 +713,17 @@ contract LiquidityBridgeContract is Initializable, OwnableUpgradeable {
             "LBC: Block height overflown"
         );
         require(
-            processedPegOutQuotes[quoteHash] != 2,
+            processedPegOutQuotes[quoteHash].statusCode != 2,
```
MaximStanciu8, line 716: maybe enum instead of magic number?

Luisfc68: Good point

MaximStanciu8 (approved):

PR #102, opened 2023-04-27
Added refund method

gsoares85:


contracts/LiquidityBridgeContract.sol
```diff
@@ -687,7 +729,10 @@ contract LiquidityBridgeContract is Initializable, OwnableUpgradeable {
         return transferredAmountOrErrorCode;
     }
 
-    function depositPegout(bytes32 quoteHash, address lpAddress) external payable {
+    function depositPegout(
+        bytes32 quoteHash,
+        address lpAddress
+    ) external payable {
```
Luisfc68, line 735: maybe its a good idea to require here that quote has not been refunded

Luisfc68 (approved):

PR #103, opened 2023-04-27
GBI-1223

Dominikkq:


MaximStanciu8 (approved):

contracts/LiquidityBridgeContract.sol
```diff
@@ -11,535 +11,522 @@ import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
...
+  function register(
+    string memory _name,
+    uint _fee,
+    uint _quoteExpiration,
+    uint _acceptedQuoteExpiration,
+    uint _minTransactionValue,
+    uint _maxTransactionValue,
+    string memory _apiBaseUrl,
+    bool _status,
+    string memory _providerType
```
Luisfc68, line 293: maybe its better to add this field as an enum with {PEGIN, PEGOUT, BOTH} because if its a normal string any typo would generate a 4th type of provider

Luisfc68 (approved):

contracts/LiquidityBridgeContract.sol
Dominikkq: Thanks, but instead of enum found that string comparison would be more straightforward, since we would still have to compare the enum types to strings later. Updating the PR now.
`require(
      keccak256(abi.encodePacked(_providerType)) ==
        keccak256(abi.encodePacked("pegin")) ||
        keccak256(abi.encodePacked(_providerType)) ==
        keccak256(abi.encodePacked("pegout")) ||
        keccak256(abi.encodePacked(_providerType)) ==
        keccak256(abi.encodePacked("both")),
      "Invalid provider type"
    );`

MaximStanciu8 (approved):

PR #105, opened 2023-05-08
Add new requires to refundUserPegout

coolwather:


Luisfc68 (approved):
We need to add a linter in LBC to avoid format changes because they add a lot of noise in the pr and its hard to find the actual change

PR #120, opened 2023-05-25
Feature/GBI-1272 - Temporary code size reduction

Luisfc68:
Hey guys so unfortunately I couldn't reduce the code size by separating the storage into a parent contract. Here is the list the things that I tried:
- Moving contract storage into a parent contract: this was the original idea that we talked on DM. But it didn't worked because the code of the parent becomes part of the contract code during compilation so it still reaches the limit
- Moving contract storage into a library and load it using unstructured storage: this one worked. The problem here is that turns out that the LBC storage is not that big, so by removing it we barely fit the limit (I think that with this approach the resulting size was 23.94Kb). Other problem is that with this method we need to load the storage in every function using, for example, `LbcStorage.getStorage()` or something like that. The problem with this is that in big functions such as registerPegin it loads too much data in the stack and the compilations fails with `Stack too deep. Try removing local variables.`
- Eliminating duplicated code: it didn't worked because it isn't too much duplicated code so the reduction is not big enough

The solution that I ended using was to separate the quote structures and quote related functions into a library. Now it compiles, it isn't a great reduction but its enough to unblock the deploy. We need to start thinking how to divide the contract in N independent modules and proxy them with the [EIP-2535](https://eips.ethereum.org/EIPS/eip-2535) because the real problem is the size of the functions so solutions based on splitting the storage won't be much effective.

If you guys identify some other function group that can be separated into a library it would also help but remember that in order to not load the library as part of the contract at compilation time, it must not  have internal functions. Anyway these are temporary measures only because if the code keeps growing we will need to use the diamond eip sooner than later.

PS: I added a truffle plugin to monitor contract size so we can measure the effectivity of the changes

### DIFF
--- a/contracts/LiquidityBridgeContract.sol
+++ b/contracts/LiquidityBridgeContract.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.8.3;
 pragma experimental ABIEncoderV2;
 
 import "./Bridge.sol";
+import "./Quotes.sol";
 import "./SignatureValidator.sol";
 import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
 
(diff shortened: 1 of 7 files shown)


Luisfc68:
I'll make a new PR to feature/security-audit-fix

PR #127, opened 2023-05-31
VULN-328 - Redesign shouldPenalizePegoutLP

Luisfc68:
Explanation of the new penalization conditions:

- "do not penalize if deposit amount is insufficient": this one was removed because amount is validated on depositPegout
- "do not penalize if deposit was not made on time": this one was changed because now we can store the deposit time during deposit pegout
- "penalize if call was not made": this one was removed because like in the security document says, there isn't really a call on pegout, also, in order to call refundPegout LP will be required to send proof of btc tx which means that transfer is always done
- "penalize if the transfer was not made on time": this one was changed because we can't know the exact time of the btc tx, but we can know the time when btc block was mined so we check if that time is above the deposit time plus the transfer time plus a btcBlockTime (a time that we give the tx to be mined, should be the average mining time of bitcoin)

### DIFF
--- a/contracts/BridgeMock.sol
+++ b/contracts/BridgeMock.sol
@@ -7,6 +7,7 @@ contract BridgeMock is Bridge {
 
     mapping(bytes32 => uint256) private amounts;
     mapping(uint256 => bytes) private headers;
+    mapping (bytes32 => bytes) private headersByHash;
 
     receive() external payable override {}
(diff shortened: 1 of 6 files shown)


Luisfc68:
https://rsklabs.atlassian.net/browse/GBI-1281

MaximStanciu8 (approved):

coolwather (approved):

PR #137, opened 2023-06-08
Index PegoutDeposit

Dominikkq:


MaximStanciu8 (approved):

contracts/LiquidityBridgeContract.sol
```diff
@@ -94,8 +94,9 @@ contract LiquidityBridgeContract is Initializable, OwnableUpgradeable {
     event PegOutRefunded(bytes32 quoteHash);
     event PegOutDeposit(
         bytes32 quoteHash,
-        uint256 amount,
-        uint256 timestamp
+        address indexed sender,
+        uint256 indexed amount,
```
Luisfc68, line 98: why all these fields as indexed if you are only going to filter by sender

PR #148, opened 2023-06-22
Feature/GBI-1285 - VULN-326

Luisfc68:


test/basic.tests.js
```diff
@@ -464,7 +465,7 @@ contract("LiquidityBridgeContract", async (accounts) => {
   it("should fail on contract call due to invalid user btc refund address", async () => {
     let rskRefundAddress = accounts[2];
     let destAddr = mock.address;
-    let data = web3.eth.abi.encodeFunctionCall(mock.abi[0], ["12"]);
```
MaximStanciu8, line 467: what is the magic number?

test/basic.tests.js
```diff
@@ -1092,8 +1100,11 @@ contract("LiquidityBridgeContract", async (accounts) => {
       1000
     );
 
+    const quoteHash = await instance.hashPegoutQuote(quote);
+    const signature = await web3.eth.sign(quoteHash, liquidityProviderRskAddress);
     const tx = instance.depositPegout(
       quote,
+      signature,
       { value: 500 }
```
MaximStanciu8, line 1108: and what is 500?

test/basic.tests.js
```diff
@@ -1107,9 +1118,13 @@ contract("LiquidityBridgeContract", async (accounts) => {
...
       1000
     );
+
+    const quoteHash = await instance.hashPegoutQuote(quoteExpiredByBlocks);
+    const signature = await web3.eth.sign(quoteHash, liquidityProviderRskAddress);
     quoteExpiredByBlocks.expireBlock = await web3.eth.getBlock("latest").then(block => block.number - 1);
     const revertByBlocks = instance.depositPegout(
       quoteExpiredByBlocks,
+      signature,
       { value: 2000 }
```
MaximStanciu8, line 1128: same, meaning of 2000?

test/miscellaneous.tests.js
```diff
@@ -298,7 +299,7 @@ contract("LiquidityBridgeContract", async (accounts) => {
     let destAddr = accounts[1];
     let fedBtcAddress = "0x0000000000000000000000000000000000000000";
     let liquidityProviderRskAddress = accounts[0];
-    let data = web3.eth.abi.encodeFunctionCall(mock.abi[0], [
+    let data = web3.eth.abi.encodeFunctionCall(mock.abi[1], [
       "100100001110010",
```
MaximStanciu8, line 303: what is the string?

test/basic.tests.js
Luisfc68: I don't know, all the mock.abi[0] changed to mock.abi[1] due a library update because the first position of the abi changed to another method, I just updated the indexes so I'm not sure the meaning of the value

test/basic.tests.js
Luisfc68: is the value to deposit, I will change to a variable instead

test/basic.tests.js
Luisfc68: will change to a variable too

test/miscellaneous.tests.js
Luisfc68: same situation of first comment, I don't know the meaning of the value, just updated the index

MaximStanciu8 (approved):

PR #156, opened 2023-07-19
Added require in LP register

gsoares85:


MaximStanciu8 (approved):

Luisfc68 (approved):
With this change we need to change LPS so the provider is not registered every time LPS restarts

PR #157, opened 2023-07-20
Feature/gbi 1290 - Contract split

Luisfc68:
Pls review this PR carefully. Tomorrow I'll provide more exact metrics about gas increase. Next step is update abi in lps and sdk


MaximStanciu8 (changes requested):
Select acnother target branch

PR #229, opened 2024-04-17
feat: add lp url update capability

Luisfc68:
## What
A function to allow a LP to be able to update his name and LPS URL in the contract 

## Why
So we can update URL according to the company instructions

## Task
https://rsklabs.atlassian.net/browse/GBI-1770

### DIFF
--- a/contracts/LiquidityBridgeContractV2.sol
+++ b/contracts/LiquidityBridgeContractV2.sol
@@ -88,6 +88,7 @@ contract LiquidityBridgeContractV2 is Initializable, OwnableUpgradeable, Reentra
         address userAddress
     );
     event DaoFeeSent(bytes32 indexed quoteHash, uint256 amount);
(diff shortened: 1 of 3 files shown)


contracts/LiquidityBridgeContractV2.sol
```diff
@@ -959,4 +954,19 @@ contract LiquidityBridgeContractV2 is Initializable, OwnableUpgradeable, Reentra
             emit DaoFeeSent(quoteHash, amount);
         }
     }
+
+    function updateProvider(string memory _name, string memory _url) external {
+        require(bytes(_name).length > 0 && bytes(_url).length > 0, "LBC076");
+        LiquidityProvider storage lp;
+        for (uint i = 1; i <= providerId; i++) {
```
Dominikkq, line 961: Why starting from 1 not 0?

Dominikkq (approved):

PR #230, opened 2024-05-17
Bugfix/GBI-1873 - Merge withdraw collateral functions

Luisfc68:
## What
- Merge withdraw collateral functions
- Update `BtcUtils` lib version

## Why
To prevent the bug when a LP resigned, then withdraws the pegin collateral and after that is unable to withdraw the pegout collateral

## Task
https://rsklabs.atlassian.net/browse/GBI-1873

### DIFF
--- a/contracts/LiquidityBridgeContractV2.sol
+++ b/contracts/LiquidityBridgeContractV2.sol
@@ -708,7 +708,10 @@ contract LiquidityBridgeContractV2 is Initializable, OwnableUpgradeable, Reentra
             "LBC049"
         );
         require(quote.value <= outputs[PAY_TO_ADDRESS_OUTPUT].value * (10**10), "LBC067"); // satoshi to wei
-        bytes memory btcTxDestination = BtcUtils.parsePayToPubKeyHash(outputs[PAY_TO_ADDRESS_OUTPUT].pkScript, mainnet);
+        bytes memory btcTxDestination = BtcUtils.outputScriptToAddress(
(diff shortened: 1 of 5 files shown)


contracts/LiquidityBridgeContractV2.sol
```diff
@@ -58,7 +58,6 @@ contract LiquidityBridgeContractV2 is Initializable, OwnableUpgradeable, Reentra
     event PegoutCollateralIncrease(address from, uint256 amount);
```
gsoares85, line 58: As it's a new implementation, even that is jus a small fix, and the V2 is deployed, we should create a V2_1 file and do the changes over there.

Luisfc68: I don't agree with that point because of the following reasons:
- We did it for V1 and it only introduced noise, PRs were harder to review both for us and for external teams (security) since the real changes were hidden among the rest of the code and git didn't provide any way to identify the changes since its a new file
- We're not maintaining several versions of the protocol at the same time. So keeping older versions only increases the number of files on the repo that we need to take care about
- If for some reason we need to check older versions, we are already using git as versioning tool. If you want to check the deployed code of the latest version you can refer to [this tag](https://github.com/rsksmart/liquidity-bridge-contract/releases/tag/v1.1.2)
- The versioning of the files is not accurate, we're on LBC V2 however there is not breaking change between the V1 and V2, only minor increments

Which would be the rationale for generating new files for each version? Because I don't mind doing it but I need a reason since that approach didn't bring value in the last deploy

gsoares85 (approved):

PR #232, opened 2024-06-24
Feature/GBI-1961 - Update getProviders

Luisfc68:
## What
- Update get providers function to remove the id array parameter
- Add getProvider function

## Why
This function had multiple issues when listing disabled or resigned providers, also, it required the caller to know which ids to fetch but the caller was using this function to know the available ids in first place

## Task
https://rsklabs.atlassian.net/browse/GBI-1961

### DIFF
--- a/contracts/LiquidityBridgeContractV2.sol
+++ b/contracts/LiquidityBridgeContractV2.sol
@@ -255,26 +255,37 @@ contract LiquidityBridgeContractV2 is Initializable, OwnableUpgradeable, Reentra
         return (providerId);
     }
 
-    function getProviders(
-        uint[] memory providerIds
-    ) external view returns (LiquidityProvider[] memory) {
-        LiquidityProvider[] memory providersToReturn = new LiquidityProvider[](
-            providerIds.length
-        );
+    function getProviders() external view returns (LiquidityProvider[] memory) {
         uint count = 0;
-
-        for (uint i = 0; i < providerIds.length; i++) {
-            uint id = providerIds[i];
-            if (
-                (isRegistered(liquidityProviders[id].provider) ||
-                    isRegisteredForPegout(liquidityProviders[id].provider)) &&
-                liquidityProviders[id].status
-            ) {
-                providersToReturn[count] = liquidityProviders[id];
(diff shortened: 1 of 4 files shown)


contracts/LiquidityBridgeContractV2.sol
```diff
@@ -255,28 +255,39 @@ contract LiquidityBridgeContractV2 is Initializable, OwnableUpgradeable, Reentra
...
-            uint id = providerIds[i];
-            if (
-                (isRegistered(liquidityProviders[id].provider) ||
-                    isRegisteredForPegout(liquidityProviders[id].provider)) &&
-                liquidityProviders[id].status
-            ) {
-                providersToReturn[count] = liquidityProviders[id];
+        LiquidityProvider storage lp;
+        for (uint i = 1; i <= providerId; i++) {
+            if (shouldBeListed(liquidityProviders[i])) {
```
MaximStanciu8, line 262: Help me please to understand this method

Luisfc68: Not sure if you asked about `shouldBeListed` or `getProviders` but I'll explain both:
- `shouldBeListed`: returns true if the LP should be included in the list of available LPs, this means is registered for pegin or pegout (meaning that has enough locked collateral and hasn't resigned) and has enabled status
- `getProviders`: this function returns the array of available LPs, since the `push` method is not available on memory arrays, we need to know the size of the list first, therefore, first we count how many providers accomplish the condition to be enabled and then we create the array and populate it

MaximStanciu8: Noted, i was referring to the case with lp enforcement initiative which should return the list for a specific scope/partner

MaximStanciu8 (approved):

contracts/LiquidityBridgeContractV2.sol
Luisfc68: ahhh, about LP enforcement initiative, IMO it should be handle off chain, we can discuss further once we take the epic into design phase

gsoares85 (approved):

PR #317, opened 2025-02-07
feat: extended multisig change unit tests

gsoares85:
#### **Overview**
This PR introduces an enhancement to the existing multisig integration, primarily for Gnosis Safe wallets. The following updates are included:
1. **Utility Contract (Encoder)**:
A new utility contract, `Encoder.sol`, has been added to streamline bytecode encoding for function calls. This contract simplifies the generation of ABI-encoded function calls for:
    - Setting provider status.
    - Upgrading proxies (e.g., `LiquidityBridgeContract` proxy).

2. **Extended E2E Test Cases**:
End-to-end (E2E) tests for `multisig-migration` now include:
    - Validation of multisig ownership change on the `LiquidityBridgeContract`.
    - Execution of specific contract actions (`setProviderStatus()` and proxy upgrades) via a Gnosis Safe multisig wallet.
    - Enhanced debugging with more granular logs to track the progress and identify issues.

3. **Multisig Execution Utils**:
Utility functions for executing Gnosis Safe multisig transactions (`multisigExecProviderStatusChangeTransaction` and `multisigExecUpgradeTransaction`) have been implemented:
    - These functions handle transaction hashing, signature generation, and final safe execution.
    - Supports impersonation of owners and ensures compatibility with test networks (e.g., enabling `hardhat_impersonateAccount`).

4. **Updated Migration Script**:
Minor refinements to `change-multisig-owner.ts`, ensuring consistent terminology for multisig wallet integration and improving inline documentation.

### DIFF
--- a/contracts/safe/util/Encoder.sol
+++ b/contracts/safe/util/Encoder.sol
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity >=0.7.0 <0.9.0;
+
+contract Encoder {
+    function getCallDataSetProviderStatus(uint _index, bool _status) external pure returns (bytes memory) {
+        return abi.encodeWithSignature(
+            "setProviderStatus(uint256,bool)",
+            _index,
+            _status
+        );
+    }
+
+    function getCallDataForUpgrade(address proxy, address implementation) external pure returns (bytes memory) {
+        return abi.encodeWithSignature(
+            "upgrade(address,address)",
+            proxy,
+            implementation
+        );
+    }
+}
--- a/e2e/multisig-migration.test.ts
+++ b/e2e/multisig-migration.test.ts
@@ -17,30 +17,28 @@ const { FORK_NETWORK_NAME } = process.env;
 
 const multsigInfo: MultisigInfo = multsigInfoJson;
 
-describe("Should change LBC owner to the multisig", function () {
+describe("Should change LBC owner to the multisig.ts", function () {
   it("Should change the owner", async () => {
     await checkForkedNetwork();
-
     const networkName = FORK_NETWORK_NAME ?? "rskTestnet";
+    console.log("Network name:", networkName);
 
+    const lbcName = "LiquidityBridgeContract";
     const addresses: Partial<DeploymentConfig> = read();
     const networkDeployments: Partial<DeploymentConfig[string]> | undefined =
       addresses[networkName];
-
     const lbcAddress = networkDeployments?.LiquidityBridgeContract?.address;
+    const safeAddress = multsigInfo[networkName].address;
 
     if (!lbcAddress) {
       throw new Error(
         "LiquidityBridgeContract proxy deployment info not found"
       );
     }
+    console.info(`LBC address: ${lbcAddress}`);
+    console.info(`Safe address: ${safeAddress}`);
 
-    const lbc = await ethers.getContractAt(
-      "LiquidityBridgeContractV2",
-      lbcAddress
-    );
-
-    const safeAddress = multsigInfo[networkName].address;
+    const lbc = await ethers.getContractAt(lbcName, lbcAddress);
 
     const lbcOwner = await lbc.owner();
     console.info("LBC owner:", lbcOwner);
@@ -52,6 +50,17 @@ describe("Should change LBC owner to the multisig", function () {
     ).to.not.be.reverted;
     const newLbcOwner = await lbc.owner();
     console.info("New LBC owner:", newLbcOwner);
+
+    await expect(
+      lbc.connect(impersonatedSigner).setProviderStatus(1, false)
+    ).to.be.revertedWith("LBC005");
+
+    expect(
+      multisigExecProviderStatusChangeTransaction(safeAddress, lbcAddress)
+    ).to.eventually.be.equal(true);
+    expect(
+      multisigExecUpgradeTransaction(safeAddress, lbcAddress)
+    ).to.eventually.be.equal(true);
   });
 });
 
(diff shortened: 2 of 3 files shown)


e2e/multisig-migration.test.ts
```diff
@@ -62,3 +71,229 @@ async function checkForkedNetwork() {
...
+  const safeContract = await ethers.getContractAt("GnosisSafe", safeAddress);
+
+  const EncoderFactory = await ethers.getContractFactory("Encoder");
+  const encoderDeployed = await EncoderFactory.deploy();
+  const encoder = await ethers.getContractAt(
+    "Encoder",
+    await encoderDeployed.getAddress()
+  );
+
+  const callData = await encoder.getCallDataSetProviderStatus(1, false);
```
Luisfc68, line 101: why creating a full contract for this? I think you can just use ethers interface class https://docs.ethers.org/v6/api/abi/#Interface-encodeFunctionData

contracts/safe/util/Encoder.sol
Luisfc68, line 1: this can be replaced with existing features from ethers

e2e/multisig-migration.test.ts
```diff
@@ -17,30 +17,28 @@ const { FORK_NETWORK_NAME } = process.env;
 
 const multsigInfo: MultisigInfo = multsigInfoJson;
 
-describe("Should change LBC owner to the multisig", function () {
+describe("Should change LBC owner to the multisig.ts", function () {
   it("Should change the owner", async () => {
     await checkForkedNetwork();
-
     const networkName = FORK_NETWORK_NAME ?? "rskTestnet";
+    console.log("Network name:", networkName);
```
Luisfc68, line 24:
```suggestion
    console.info("Network name:", networkName);
```

e2e/multisig-migration.test.ts
```diff
@@ -52,6 +50,17 @@ describe("Should change LBC owner to the multisig", function () {
     ).to.not.be.reverted;
     const newLbcOwner = await lbc.owner();
     console.info("New LBC owner:", newLbcOwner);
+
+    await expect(
+      lbc.connect(impersonatedSigner).setProviderStatus(1, false)
+    ).to.be.revertedWith("LBC005");
+
+    expect(
```
Luisfc68, line 58:
```suggestion
    await expect(
```

e2e/multisig-migration.test.ts
```diff
@@ -52,6 +50,17 @@ describe("Should change LBC owner to the multisig", function () {
...
+      multisigExecProviderStatusChangeTransaction(safeAddress, lbcAddress)
+    ).to.eventually.be.equal(true);
+    expect(
```
Luisfc68, line 61:
```suggestion
    await expect(
```

e2e/multisig-migration.test.ts
Luisfc68, line 1: Two things I think need to be added:
- try to upgrade using the old owner and verify it fails (right now the test only does this with the `setProviderStatus`)
- make some call after upgrading with the multisig to verify it was actually upgraded

Luisfc68 (changes requested):
The PR is not passing the linter (we need to update the action file to fail in this cases)
image

Update
Ignore this, was related to my local env

e2e/multisig-migration.test.ts
gsoares85: Solved

contracts/safe/util/Encoder.sol
gsoares85: Solved

e2e/multisig-migration.test.ts
gsoares85: Solved

e2e/multisig-migration.test.ts
gsoares85: This is using the eventually, so there is no need of await.

e2e/multisig-migration.test.ts
gsoares85: Same

e2e/multisig-migration.test.ts
gsoares85: Done 
`expect(
      multisigExecUpgradeTransaction(impersonatedSigner.address, lbc)
    ).to.eventually.be.equal(false);
    expect(
      multisigExecUpgradeTransaction(safeAddress, lbc)
    ).to.eventually.be.equal(true);`
    First call try to upgrade with the old owner and second with the new one

e2e/multisig-migration.test.ts
Luisfc68: same answer

e2e/multisig-migration.test.ts
```diff
@@ -52,6 +51,20 @@ describe("Should change LBC owner to the multisig", function () {
...
+
+    await expect(
+      lbc.connect(impersonatedSigner).setProviderStatus(1, false)
+    ).to.be.revertedWith("LBC005");
+
+    expect(
+      multisigExecProviderStatusChangeTransaction(safeAddress, lbc)
+    ).to.eventually.be.equal(true);
+    expect(
+      multisigExecUpgradeTransaction(impersonatedSigner.address, lbc)
```
Luisfc68, line 63: I have a doubt about this line, if your impersonated signer is an EOA, why are you passing it as it was a multisig address? I think you should try to just upgrade like you did in line 56 with `setProviderStatus`

e2e/multisig-migration.test.ts
```diff
@@ -52,6 +51,20 @@ describe("Should change LBC owner to the multisig", function () {
...
+    ).to.eventually.be.equal(false);
+    expect(
```
Luisfc68, line 65: same comment about the await expect

e2e/multisig-migration.test.ts
Luisfc68: I think first call is incorrect, I left a separated comment explaining why

Luisfc68 (changes requested):

Luisfc68 (approved):

PR #348, opened 2025-07-28
Split preparations

Luisfc68:
## What
- Update `solhint` version and configuration
- Fix the lint issues found by the new linter version
- Organize repository structure

## Why
After doing some PoCs with the gas consumption of the discovery part of the protocol, we need to prepare the repo to start receiving the actual implementation of the split. This includes organizing the current contract files into a clear repository structure and update the linter so the new contributions can be consistent between them.

## Important
The new directory structure is the following:
.
└── contracts/
    ├── interfaces/
    ├── legacy/
    ├── libraries/
    ├── test-contracts/
    ├── split/
    ├── Contract1
    ├── Contract2
    ├── ...
    └── ContractN
where the directories will be used in the following way:
- **interfaces**: this will contain the interface of each one of the contracts, this directory is very important as the contracts will interact between them using interfaces and the off-chain components will generate their bindings from this interfaces too.
- **legacy**: this contains the files of the non-split LBC version, once the new version is validated and confirmed this dir can be removed.
- **libraries**: this will only contain libraries so we can keep them separated from the contracts.
- **test-contracts**: smart contracts used for unit and integration tests.
- **split**: files that are being used for the sprint but are not yet part of the final version. E.g. the files we used for the Discovery gas measurements. This dir should disappear at the end of the split implementation too.
- **contracts**: the main contracts of the protocol (PegIn, PegOut, Discovery, CollateralManagement)

## Not included in this PR
- NatSpec documentation, as it will be provided with the implementation of each one of the contracts
- Readme update with the directory structure, as we need to create a separate task for the overall documentation of the repo

## Task
https://rsklabs.atlassian.net/browse/GBI-2842


alexjavabraz (approved):
LGTM

Dominikkq (approved):

Hakob23:
NIT: We could also rename the file names of the interfaces to follow the IName.sol convention. Otherwise, LGTM!

PR #353, opened 2025-08-11
Adds ghr workflow

rafaiovlabs:


.github/workflows/docker.yml
```diff
@@ -0,0 +1,76 @@
+name: Build and Push Docker Image
+
+on:
+  push:
```
Luisfc68, line 4: Maybe we can publish only with the releases and the workflow_dispatch, I think the PR and the push are not necessary because we do many of them before getting to the stable releases, WDYT?

alexjavabraz: My suggestion is to publish only when any v* tag is created

rafaiovlabs: Whatever you need as the workflow is working now it can be modified by the team

rafaiovlabs (commented):
Can this be merged ?

PR #354, opened 2025-08-21
Feature/GBI-2807 - PegIn contract

Luisfc68:
## What
- Add pegin contract
- Add pegin contract documentation
- Add unit tests for all the pegin contract functions

## Why
To continue with the LBC split, the remaining contracts are `CollateralManagement` and `FlyoverDiscovery`

## Task
https://rsklabs.atlassian.net/browse/GBI-2807

## Important
The punisher reward mechanism was moved to `CollateralManagement` instead of the `PegInContract`. The reason of this is because the tokens used to reward the punisher (whomever calls the `registerPegIn` to penalize a liquidity provider) are taken from the provider's collateral, so it makes more sense to have that mechanism in `CollateralManagement`. The notion page will be updated with this change.


contracts/PegInContract.sol
```diff
@@ -0,0 +1,539 @@
...
+            msg.sender,
+            quote.contractAddress,
+            quoteHash,
+            quote.gasLimit,
+            quote.value,
+            quote.data,
+            success
+        );
+        return success;
+    }
```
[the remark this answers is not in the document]
alexjavabraz: Please evaluate the usage of Checks-Effects-Interactions pattern here.

alexjavabraz (changes requested):
Evaluate the usage of Checks-Effects-Interactions pattern

hakobRsk (commented):
Is there a reason, the `getMinPegin`  is not implemented(it is mentioned in the docs but not present in the legacy contract)? 
And I assume `getRewardPercentage` and `rewardP` is not implemented here as they are the part of the punishing mechanism, right?

Otherwise, LGTM

contracts/PegInContract.sol
```diff
@@ -0,0 +1,539 @@
...
+        _increaseBalance(quote.liquidityProviderRskAddress, msg.value);
+
+        // This check ensures that the call cannot be performed with less gas than the agreed amount
+        if (gasleft() < quote.gasLimit + _MAX_CALL_GAS_COST) {
+            revert InsufficientGas(gasleft(), quote.gasLimit + _MAX_CALL_GAS_COST);
+        }
+        _callRegistry[quoteHash].timestamp = block.timestamp;
+        _processedQuotes[quoteHash] = PegInStates.CALL_DONE;
+
+        (bool success,) = quote.contractAddress.call{
```
hakobRsk, line 180: 1. Currently we don't have a behavior to revert if the call(in any way) returns `false` . May be we should handle that case inside the contract as well?
2. This part is also flagged as reentrancy vulnerability. If we agree to add a revert mechanism as mentioned in the first point, then maybe we could do smth like this?
```
_callRegistry[quoteHash].success = true;
_decreaseBalance(quote.liquidityProviderRskAddress, quote.value);
 
  
 (bool success,) = quote.contractAddress.call{
    gas: quote.gasLimit,
    value: quote.value
}(quote.data);
        
if(!success) revert();

Comment (Luisfc68) [reply contracts/PegInContract.sol:180]: Both questions are related, the context here is that we need to keep track of the result of the call to the `contractAddress`, as it affects the flow in the following ways:

1. If the LP **did** make the call, but the call wasn't successful, then they aren't applicable for penalization (assuming timing was ok) and they are still refunded with the `gasFee` and `callFee`, but not the value as the call reverted
2. If the LP **did** make the call and was successful, then they aren't applicable for penalization (assuming timing was ok) and they are refunded with `gasFee`, `callFee` and `value` amounts
3. If the LP **didn't** make the call then they are applicable for penalization (assuming timing was ok) but in that case the `callForUser` never executes 

The point is that if we don't keep track of the result of the call in the state of the contract then we won't be to determine most of the scenarios inside the `registerPegIn`. This only happens with the `CallForUser` event, so if you see something similar in other part of the contract then we should change that but in this particular case I think we are pretty much forced to maintain the modification of the state after the call. Anyway if you have any proposal to rewrite the function but maintaining the overall logic we can try with that, perhaps we could add 
```
_callRegistry[quoteHash].success = true;
_decreaseBalance(quote.liquidityProviderRskAddress, quote.value);
```
optimistically before the external call, but we'd still need to emit the event with the result of the call and rollback that modification in case the call fails, so we would be in a similar situation.

Regarding the reentrancy alert, yes, even though I included some [tests](https://github.com/rsksmart/liquidity-bridge-contract/pull/354/files#diff-74eecfcc2f6b6b3cd120d85d03eff12a580dd9c3b96d6dd398d4c83e627c85daR389) specifically to ensure this shouldn't be affected by that vulnerability, slither seems not to be detecting the `nonReentrant` modifier. I already raised this topic with security team and I'm waiting for their response.

Luisfc68:
> Is there a reason, the `getMinPegin` is not implemented(it is mentioned in the docs but not present in the legacy contract)? And I assume `getRewardPercentage` and `rewardP` is not implemented here as they are the part of the punishing mechanism, right?
> 
> Otherwise, LGTM

@hakobRsk There is no reason, I forgot to add that getter, thanks for noticing.

Regarding `getRewardPercentage` correct, as explained in the PR description that mechanism is being moved to Collateral Management so those variables are going with it

contracts/PegInContract.sol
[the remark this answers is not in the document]
Luisfc68: Hakob made a similar comment related to the reentrancy alert, I explained the rationale of doing it in this way in my response, please take a look a let me know your thoughts about the need of keeping track of the result of the call.

The response -> https://github.com/rsksmart/liquidity-bridge-contract/pull/354#discussion_r2297567027

contracts/PegInContract.sol
hakobRsk: Okay, thanks for the clarification. I think we can keep it this way then, especially if there is an explicit reentrancy guard. We could still try to follow the DEI pattern, but that would unnecessarily add the gas cost for the caller.

Hakob23 (approved):

alexjavabraz (approved):
LGTM

PR #356, opened 2025-08-27
Gbi 2845/discovery contract implementation

Hakob23:
### What

- Add Discovery contract
- Add unit tests and docs

### Why

To finalize with the implementation of the split contracts

Task
https://rsklabs.atlassian.net/browse/GBI-2845


contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,266 @@
...
+    // ------------------------------------------------------------
+
+    mapping(uint => Flyover.LiquidityProvider) private _liquidityProviders;
+    ICollateralManagement private _collateralManagement;
+    uint public lastProviderId;
+
+    // ------------------------------------------------------------
+    // Basic Configuration State Variables
+    // ------------------------------------------------------------
+
```
Luisfc68: these two can be removed, in the case of `_minCollateral` it belongs to `CollateralManagement`, in the case of `_resignationBlockNum` even though we planned to include it here, it seems to make more sense to have it in `CollateralManagement` as well, first because the resign mechanism is used to tell if the LP can withdraw the collateral or not, and second because if this contract needs to check if the LP resigned or not for the `isOperational` function that can be easily fetched from the current `CollateralManagement` implementation

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,266 @@
...
+            providerType: providerType
+        });
+        emit Register(lastProviderId, msg.sender, msg.value);
+        return (lastProviderId);
+    }
+
+    /// @notice Resigns the caller as a Liquidity Provider
+    /// @dev Reverts if the caller is not registered or already resigned
+    /// @dev Resignation is permanent and cannot be undone
+    function resign() external override {
```
Luisfc68: by the same rationale explained in https://github.com/rsksmart/liquidity-bridge-contract/pull/356/files#r2303909241 this can be removed from discovery now

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,266 @@
...
+            _collateralManagement.getPegInCollateral(providerAddress) <= 0 &&
+            _collateralManagement.getPegOutCollateral(providerAddress) <= 0
+        ) {
+            revert Flyover.ProviderNotRegistered(providerAddress);
+        }
+        _resignationBlockNum[providerAddress] = block.number;
+        emit Resigned(providerAddress);
+    }
+
+    // non-view external functions should be declared before external view functions (solhint ordering)
```
Luisfc68:
```suggestion
```

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,266 @@
...
+            status: status,
+            providerType: providerType
+        });
+        emit Register(lastProviderId, msg.sender, msg.value);
+        return (lastProviderId);
+    }
+
+    /// @notice Resigns the caller as a Liquidity Provider
+    /// @dev Reverts if the caller is not registered or already resigned
+    /// @dev Resignation is permanent and cannot be undone
```
Luisfc68:
```suggestion
    /// @dev Resignation is permanent and cannot be undone. To recover the collateral, the liquidity provider must go through the resignation process, if that happens and the same account wishes to register again, that would result in a different provider ID. 
```

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,266 @@
...
+                return;
+            }
+        }
+        revert Flyover.ProviderNotRegistered(providerAddress);
+    }
+
+        /// @notice Sets the minimum collateral threshold per service side
+        /// @dev Only callable by DEFAULT_ADMIN_ROLE
+        /// @param minCollateral New minimum collateral value
+        function setMinCollateral(uint minCollateral) external onlyRole(DEFAULT_ADMIN_ROLE) {
```
Luisfc68: this is part of `CollateralManagement`, it can be removed from this contract

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,266 @@
...
+        _minCollateral = minCollateral;
+    }
+
+    /// @notice Lists LPs that should be visible to users
+    /// @dev A provider is listed if it has sufficient collateral for at least one side and `status` is true
```
Luisfc68:
```suggestion
    /// @dev A provider is listed if they have collateral for the operation they're registered for and `status` is true
```

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,266 @@
...
+    /// @return providersToReturn Array of LP records to display
+    function getProviders() external view returns (Flyover.LiquidityProvider[] memory) {
```
Luisfc68, line 102: This implementation is the same as in the legacy version so it is ok, however, now that `_shouldBeListed` involves an external call we might want to reduce the usage of that function to only once, this the proposal:

```
function getProviders() external view returns (Flyover.LiquidityProvider[] memory) {
        uint256[] memory matchingLps = new uint256[](lastProviderId);
        uint count = 0;
        Flyover.LiquidityProvider storage lp;
        for (uint i = 1; i <= lastProviderId; i++) {
            lp = _liquidityProviders[i];
            if (_shouldBeListed(lp)) {
                matchingLps[count] = lp.id;
                count++;
            }
        }
        Flyover.LiquidityProvider[] memory providers = new Flyover.LiquidityProvider[](count);
        for (uint i = 0; i < count; i++) {
            providers[i] = _liquidityProviders[matchingLps[i]];
        }
        return providers;
    }

    function _shouldBeListed(Flyover.LiquidityProvider storage lp) private view returns(bool){
        return lp.status && collateralManagement.isRegistered(lp.providerType, lp.providerAddress);
    }
```

(is not tested is just one proposal to implement the same but reducing the external calls, is taken from the `FlyoverDiscoveryContract` version used in the benchmarks with some modifications)

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,266 @@
...
+    /// @return provider LP record, reverts if not found
+    function getProvider(address providerAddress) external view returns (Flyover.LiquidityProvider memory) {
+        return _getProvider(providerAddress);
+    }
+
+    /// @notice Checks if an LP can operate for peg-in side
+    /// @dev Ignores the first argument as compatibility stub with legacy signature
+    /// @param addr The LP address
+    /// @return isOp True if registered and peg-in collateral >= min
+    function isOperational(Flyover.ProviderType, address addr) external view returns (bool) {
```
Luisfc68: I think we should use the `isOperational` version used in the split PoC (the one that absorbs `isOperational` and `isOperationalForPegout` in the same function), as it is easier to consume from other contracts / off chain components

https://github.com/rsksmart/liquidity-bridge-contract/blob/lbc-split/contracts/split/FlyoverDiscovery.sol#L111

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,266 @@
...
+    function _getProvider(address providerAddress) private view returns (Flyover.LiquidityProvider memory) {
+        for (uint i = 1; i <= lastProviderId; ++i) {
+            if (_liquidityProviders[i].providerAddress == providerAddress) {
+                return _liquidityProviders[i];
+            }
+        }
+        revert Flyover.ProviderNotRegistered(providerAddress);
+    }
+
+    function _isRegistered(address addr) private view returns (bool) {
```
Luisfc68: since the resignation part is being moved to CollateralManagement then you can remove this one and the one below and call the other contract to check

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,266 @@
...
+    /// @return resignationBlock Block number when resignation started, or 0 if not resigned
+    function getResignationBlockNum(address addr) external view returns (uint256) {
+        return _resignationBlockNum[addr];
+    }
+
+    // ------------------------------------------------------------
+    // FlyoverDiscovery Private Functions
+    // ------------------------------------------------------------
+
+    function _shouldBeListed(Flyover.LiquidityProvider storage lp) private view returns(bool){
```
Luisfc68, line 172: same here we can use the PoC version https://github.com/rsksmart/liquidity-bridge-contract/blob/lbc-split/contracts/split/FlyoverDiscovery.sol#L116

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,266 @@
...
+        if (
+            bytes(name).length < 1 ||
+            bytes(apiBaseUrl).length < 1
+        ) {
+            revert InvalidProviderData(name, apiBaseUrl);
+        }
+
+        if (providerType > type(Flyover.ProviderType).max) revert InvalidProviderType(providerType);
+
+        if (_resignationBlockNum[providerAddress] != 0) {
```
Luisfc68: this condition and the ones below should change since the minCollateral doesn't go on discovery, you can check here for reference https://github.com/rsksmart/liquidity-bridge-contract/blob/lbc-split/contracts/split/FlyoverDiscovery.sol#L138

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,266 @@
...
+    /// @param apiBaseUrl Base URL of the LP public API
+    /// @param status Initial status flag (enabled/disabled)
+    /// @param providerType The service type(s) the LP offers
+    /// @return id The newly assigned LP identifier
+    function register(
+        string calldata name,
+        string calldata apiBaseUrl,
+        bool status,
+        Flyover.ProviderType providerType
+    ) external payable returns (uint) {
```
Luisfc68, line 54: this function is not sending the collateral received to the CollateralManagement contract, that step is very important (rememer to give the COLLATERAL_ADDER role to this contract in the script to be able to do so)

Luisfc68 (commented):
Mostly looks good, there are a couple things to change:
- Move some parts to `CollateralManagement`
- Send the money received in register to `CollateralManagement`

Also, we might want to add a way for the different parties to verify which are the addresses of the other contracts, we can discuss later if this is the proper contract to do so or not

contracts/FlyoverDiscovery.sol
Hakob23: I have removed `minCollateral` check and for the `AlreadyRegistered` condition, I think we could simply use `_collateralManagement.isRegistered` function instead of what is done in the PoC.

contracts/interfaces/IFlyoverDiscovery.sol
```diff
@@ -15,16 +15,50 @@ interface IFlyoverDiscovery {
...
+    /// @dev A provider is listed if it has sufficient collateral for at least one side and `status` is true
+    /// @return providersToReturn Array of LP records to display
     function getProviders() external view returns (Flyover.LiquidityProvider[] memory);
+
+    /// @notice Returns a single LP by address
+    /// @param providerAddress The LP address
+    /// @return provider LP record, reverts if not found
     function getProvider(address providerAddress) external view returns (Flyover.LiquidityProvider memory);
+
+    /// @notice Checks if an LP can operate for peg-in side
```
Luisfc68:
```suggestion
    /// @notice Checks if an LP can operate for a specific type of operation
```

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,233 @@
...
+        ++lastProviderId;
+        _liquidityProviders[lastProviderId] = Flyover.LiquidityProvider({
+            id: lastProviderId,
+            providerAddress: msg.sender,
+            name: name,
+            apiBaseUrl: apiBaseUrl,
+            status: status,
+            providerType: providerType
+        });
+        _addCollateral(providerType, msg.sender, msg.value);
```
Luisfc68: I think you should emit `Register` before, because `_addCollateral` has an external call inside

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,233 @@
...
+    /// @param owner The Default Admin and initial owner address
+    /// @param initialDelay The initial admin delay for `AccessControlDefaultAdminRulesUpgradeable`
+    /// @param collateralManagement The address of the `ICollateralManagement` contract
+    function initialize(
+        address owner,
+        uint48 initialDelay,
+        address collateralManagement
+    ) external initializer {
+        __AccessControlDefaultAdminRules_init(initialDelay, owner);
+        _collateralManagement = ICollateralManagement(collateralManagement);
```
Luisfc68, line 45: I suggest to check if there's any code inside `collateralManagement` like in the other contracts `if (collateralManagement.code.length == 0) revert Flyover.NoContract(collateralManagement);`

contracts/FlyoverDiscovery.sol
Luisfc68: I maintain this comment from the previous review

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    /// @inheritdoc IFlyoverDiscovery
+    function getProvider(address providerAddress) external view returns (Flyover.LiquidityProvider memory) {
+        return _getProvider(providerAddress);
+    }
+
+    /// @inheritdoc IFlyoverDiscovery
+    function isOperational(Flyover.ProviderType, address addr) external view returns (bool) {
+        return _collateralManagement.isCollateralSufficient(Flyover.ProviderType.PegIn, addr) &&
```
Luisfc68: Here you should pass the 1st argument too, there's no problem, this contract has its own test suite so we don't need to keep 100% compatibility with the legacy version, when the time comes to adapt the legacy test suite we'll just replace the calls to the old functions in there. 

Additionally,  I recommend to swap the condition here, `_getProvider` doesn't have an external call so it the status happens to be false we can save the cost of checking the collateral

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,233 @@
...
+    IFlyoverDiscovery
+{
+
+    // ------------------------------------------------------------
+    // FlyoverDiscovery State Variables
+    // ------------------------------------------------------------
+
+    mapping(uint => Flyover.LiquidityProvider) private _liquidityProviders;
+    ICollateralManagement private _collateralManagement;
+    uint public lastProviderId;
```
Luisfc68, line 27: now that we have a getter perhaps this should be private

contracts/interfaces/IFlyoverDiscovery.sol
Luisfc68, line 1: Regarding the interfaces, my advice is to just leave one called `IFlyoverDiscovery.sol` for the interface of the split version and then update `ILegacy.sol` with anything you need for compatibility purposes

contracts/split/FlyoverDiscovery.sol
Luisfc68, line 1: Feel free to remove this file and `FlyoverDiscoveryFull.sol` too, since you're adding the actual implementation we don't need them anymore

test/discovery/fixtures.ts
```diff
@@ -0,0 +1,96 @@
+import { upgrades, ethers } from "hardhat";
+import { ProviderType } from "../utils/constants";
+
+export async function deployDiscoveryFixture() {
+  const CollateralManagement = await ethers.getContractFactory(
```
Luisfc68: instead of deploying `CollateralManagement` here manually I recommend to call the fixture that is inside the collateral management test dir (call it normally, not as a fixture)

test/discovery/operational.test.ts
```diff
@@ -0,0 +1,22 @@
...
+  it("isOperational returns true only for providers with peg-in collateral", async () => {
+    const { discovery, pegInLp, fullLp, pegOutLp } = await loadFixture(
+      deployDiscoveryWithProvidersFixture
+    );
+
+    expect(
+      await discovery.isOperational(ProviderType.PegIn, pegInLp.address)
+    ).to.equal(true);
+    expect(
+      await discovery.isOperational(ProviderType.PegIn, fullLp.address)
```
Luisfc68, line 17: Same comment as in the contract, this function should work using the 1st arg, don't worry about the compatibility with the legacy tests as we're not adapting them yet

test/discovery/resign.test.ts
```diff
@@ -0,0 +1,366 @@
...
+import { deployLbcWithProvidersFixture } from "../utils/fixtures";
+import {
+  createBalanceDifferenceAssertion,
+  createBalanceUpdateAssertion,
+  createCollateralUpdateAssertion,
+} from "../utils/asserts";
+
+describe("Discovery resign flow should", () => {
+  it("emit Resigned and hide provider from listing", async () => {
+    const fixtureResult = await loadFixture(deployLbcWithProvidersFixture);
```
Luisfc68: This is the fixture of the legacy contract so this tests are covering only that and not the split version (and I think the legacy suite already have these cases), the test is correct but instead of doing it with the legacy fixture you should use discovery + collateral management contracts

Luisfc68: same comment in general for the other tests in this file

test/discovery/resign.test.ts
```diff
@@ -0,0 +1,366 @@
...
+        .withArgs(lp.signer.address);
+
+      const lpBalanceAfterCollateralWithdrawAssertion =
+        await createBalanceUpdateAssertion({
+          source: ethers.provider,
+          address: lp.signer.address,
+          message: "Incorrect LP balance after collateral withdraw",
+        });
+
+      const lbcBalanceAfterCollateralWithdrawAssertion =
```
Luisfc68: Also there's no need of using this helpers anymore as hardhat has the assertions to check if a tx changed the ether balance

Luisfc68 (changes requested):

contracts/FlyoverDiscovery.sol
Hakob23: I think we can just remove the first argument then. Agreed with swapping

Luisfc68: no no is the other way around, pass the argument so we can query for the three states

contracts/FlyoverDiscovery.sol
```diff
@@ -0,0 +1,234 @@
...
+        return providersToReturn;
+    }
+
+    /// @inheritdoc IFlyoverDiscovery
+    function getProvider(address providerAddress) external view returns (Flyover.LiquidityProvider memory) {
+        return _getProvider(providerAddress);
+    }
+
+    /// @inheritdoc IFlyoverDiscovery
+    function isOperational(address addr) external view returns (bool) {
```
Luisfc68: this still don't receive the operation type

Luisfc68 (commented):
besides the isOperational comment the rest seems to be ok

Luisfc68 (approved):

PR #361, opened 2025-09-21
Gbi 2875/2 make transferred amount check strict

Hakob23:
[The Task](https://rsklabs.atlassian.net/browse/GBI-2875)

**What**

Make transferred amount check strict as recommended per audit

**Why**

If a calling contract uses `checkAgreedAmount` function to check the incoming amount (`msg.value`) and then proceeds to distribute funds based on the individual values in the quote (which sum up to `agreedAmount`), a transaction can pass the check but fail during the fund distribution phase due to insufficient balance. This creates a gas grieving vulnerability, especially in a system that uses relayers (as suggested by the presence of a `gasFee` in the quote), where a malicious user can make a relayer submit a transaction that is guaranteed to fail, causing the relayer to lose the gas fees.

**How to Test**

`npm run test test/pegin/register-pegin.test.ts`


contracts/libraries/Quotes.sol
```diff
@@ -57,11 +57,9 @@ library Quotes {
         agreedAmount = quote.value + quote.callFee + quote.productFeeAmount + quote.gasFee;
 
 
-        uint delta = agreedAmount / 10000;
-        // transferred amount should not be lower than (agreed amount - delta),
-        // where delta is intended to tackle rounding problems
-        if (agreedAmount - delta > transferredAmount) {
-            revert AmountTooLow(transferredAmount, agreedAmount - delta);
+        // transferred amount should not be lower than agreed amount
+        if (agreedAmount > transferredAmount) {
```
Luisfc68, line 68: this is ok according to the audit, but doesn't consider the rounding issues, when going from sats to wei,the precision differences between both networks might generate a non refundable quote if they aren't considered. My recommendation is to remove the delta as the report says, but implement this check in a similar way to the one we do in pegout https://github.com/rsksmart/liquidity-bridge-contract/blob/master/contracts/LiquidityBridgeContractV2.sol#L718C8-L722C88 in that way, we allow the bitcoin transaction to either round or truncate the wei value, protecting the user from cases where the client app doesn't round the wei -> sat value properly

Luisfc68 (changes requested):

Luisfc68 (approved):

PR #362, opened 2025-09-21
Gbi 2866/emergency pause

Hakob23:
[The Task](https://rsklabs.atlassian.net/jira/software/projects/GBI/boards/42?selectedIssue=GBI-2866)

**What**

Design Emergency Pause for Liquidity Provider


contracts/EmergencyPause/EmergencyPauserPeg.sol
Luisfc68, line 1: Question:
I guess the decision of implementing 2 contracts is because both peg contracts are `DaoContributor` and therefore inherit from `AccessControlDefaultAdminRulesUpgradeable`. Now, consider that is very likely that `CollateralManagement` becomes `DaoContributor` in the short-mid term, in that case do you still like this approach? If yes, ok, then maybe we should rename `EmergencyPauserPeg` and use other name, if not, perhaps we should add the access control part in a common contract for ALL the flyover contracts

Hakob23: @Luisfc68 , The main reason to have 2 separate contracts was that `AccessControlDefaultAdminRulesUpgradeable` which was used in `Discovery` and `Collateral` ,  and `OwnableDaoContributorUpgradeable` which was used in `Pegin` and `Pegout` are conflicting to each other. That's why there was a need to implement separatly `AccessControlUpgradeable` for `Pegin` and `Pegout` in order not to run into conflicts. Regarding whether `CollateralManagment` becoming `DaoContributor`: it depends on how exactly are we going to solve the conflict implementations between them. Regarding the name, agreed, we could use something better, may be `EmergencyDaoContributor`? wdyt about these points?

Luisfc68: Maybe `PausableDaoContributor` instead of `EmergencyDaoContributor`? As the pausable word is probably more explanatory. Regarding the Ownable vs AccessControl, IMO maybe it would be better to change them so they all intherit from AccessControl, anyway we already have at least 2 roles in each contract, so we can keep it extensible in that way

Hakob23: I was think of changing from ownable to AccessControl for Pegin and Pegout, but was afraid that it could get out of the scope of this PR. Are you suggesting to make these changes now?

Luisfc68: Yes, if you prefer doing it here feel free to continue with the PR, if not feel free to use a separate PR/task

PR #364, opened 2025-10-01
Gbi 2875/6 register call not done fix

Hakob23:
**What**

Add fallback mechanism is case of external call failure inside  `_registerCallNotDone` 

**Why**

To prevent funds from being locked forever

**How to test**

`npm run test ./test/pegin/register-pegin.test.ts`

### DIFF
--- a/test/pegin.test.ts
+++ b/test/pegin.test.ts
@@ -9,6 +9,7 @@ import { loadFixture } from "@nomicfoundation/hardhat-toolbox/network-helpers";
 import { expect } from "chai";
 import { deployContract } from "../scripts/deployment-utils/utils";
 import { deployLbcWithProvidersFixture } from "./utils/fixtures";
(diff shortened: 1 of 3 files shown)


test/pegin.test.ts
```diff
@@ -1760,4 +1761,631 @@ describe("LiquidityBridgeContractV2 pegin process should", () => {
...
+    );
+
+    expect(refundBalanceIncreaseEvents.length).to.eq(0); // No balance increase for refund!
+
+    const maliciousContractBalanceAfter = await lbc.getBalance(
+      maliciousContractAddress
+    );
+
+    // Verify the balance was NOT increased (funds are NOT credited)
+    expect(
```
Luisfc68, line 1890: remember you can use expect(tx).to.changeEtherBalances(...)

Luisfc68 (approved):

PR #369, opened 2025-10-28
Fly 2075/add tests and tasks in foundry

Hakob23:
**What**

[The task](https://rsklabs.atlassian.net/jira/software/c/projects/FLY/boards/682?assignee=712020%3A06f65950-a686-4803-85cf-442c880e9bcf&selectedIssue=FLY-2075&useStoredSettings=true)


forge-test/collateral/Configuration.t.sol
```diff
@@ -0,0 +1,186 @@
...
+        // Check penalties
+        assertEq(
+            collateralManagement.getPenalties(),
+            0,
+            "Penalties should be 0"
+        );
+    }
+
+    function test_Initialize_AllowsInitializeOnlyOnce() public {
+        vm.expectRevert(); // InvalidInitialization error
```
Luisfc68: maybe we could assert the type of the error here

forge-test/collateral/Configuration.t.sol
```diff
@@ -0,0 +1,186 @@
...
+            )
+        );
+        collateralManagement.setMinCollateral(1);
+    }
+
+    function test_SetMinCollateral_ModifiesProperly() public {
+        uint256 newMinCollateral = 11;
+
+        vm.prank(owner);
+        // Note: Event is emitted but we just check the state change
```
Luisfc68: any particular reason not to check the event?

forge-test/collateral/Resign.t.sol
```diff
@@ -0,0 +1,818 @@
...
+import {ICollateralManagement} from "../../contracts/interfaces/ICollateralManagement.sol";
+import {Flyover} from "../../contracts/libraries/Flyover.sol";
+import {Quotes} from "../../contracts/libraries/Quotes.sol";
+import {WalletMock} from "../../contracts/test-contracts/WalletMock.sol";
+
+contract ResignTest is CollateralTestBase {
+    address public notProvider;
+
+    // Events from ICollateralManagement
+    event Resigned(address indexed addr);
```
Luisfc68: why can't we use the interface directly?

forge-scripts/tasks/HashQuote.s.sol
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        return bytes20(sliced);
+    }
+
+    /**
+     * @notice Get LBC address from deployment config or environment variable
+     * @return The LBC contract address
+     */
+    function getLbcAddress() internal view returns (address) {
```
Luisfc68, line 73: I think is better to just enforce the env var since its less error prone, anyway if this is unit tested and working feel free to leave it as is

forge-test/tasks/HashQuote.t.sol
```diff
@@ -0,0 +1,185 @@
...
+        // Update the lbcAddress to match our test contract
+        quote.lbcAddress = address(lbc);
+
+        // Hash the quote
+        console.log("LBC address:", address(lbc));
+        bytes32 hash = lbc.hashQuote(quote);
+        console.log("Quote hash:");
+        console.logBytes32(hash);
+
+        assertTrue(hash != bytes32(0), "Hash should not be zero");
```
Luisfc68, line 85: I think you should tests with real examples, let me know if you need a dataset to test here but since the current implementation of the task is not trivial I think that testing the hash is not zero is not enough

forge-test/tasks/HashQuote.t.sol
```diff
@@ -0,0 +1,185 @@
...
+
+        // Create a test quote directly
+        QuotesV2.PeginQuote memory quote = createTestPeginQuote();
+
+        // Hash using contract directly
+        bytes32 contractHash = lbc.hashQuote(quote);
+        console.log("Hash from contract:");
+        console.logBytes32(contractHash);
+
+        // The script uses the same contract method, so hashes should match
```
Luisfc68, line 119: There is no assertion in this test

forge-test/tasks/HashQuote.t.sol
```diff
@@ -0,0 +1,185 @@
...
+
+        // Create a test pegout quote directly
+        QuotesV2.PegOutQuote memory quote = createTestPegoutQuote();
+
+        // Hash using contract directly
+        bytes32 contractHash = lbc.hashPegoutQuote(quote);
+        console.log("Hash from contract:");
+        console.logBytes32(contractHash);
+
+        // The script uses the same contract method, so hashes should match
```
Luisfc68, line 138: same here, no assertion to verify the hash produced is correct

Makefile
```diff
@@ -230,13 +287,260 @@ change-owner-broadcast:
...
+	@echo "Reason: $(PAUSE_REASON)"
+	@export NETWORK=$(call get_rsk_network_name,$(NETWORK)); \
+	forge script forge-scripts/tasks/PauseSystem.s.sol:PauseSystem \
+		--sig "pauseAll(string)" "$(PAUSE_REASON)" \
+		--rpc-url $(call get_network_config,$(NETWORK)) \
+		-vv
+
+# Pause all system contracts (actual broadcast)
+.PHONY: pause-system-broadcast
+pause-system-broadcast:
```
Luisfc68, line 354: I think we can skip the ledger usage from this scripts because if the script is important enough to involve a ledger it will generally be handled by the owner's or pauser's multisig

Luisfc68: I'm talking in general, not only for this case

forge-scripts/tasks/RefundUserPegout.s.sol
```diff
@@ -0,0 +1,457 @@
...
+                // Use default from private key if available
+                sender = vm.addr(vm.envUint("PRIVATE_KEY"));
+            }
+        }
+
+        // Estimate gas by simulating the call
+        vm.prank(sender);
+        try lbc.refundUserPegOut(quoteHash) {
+            // If we get here in simulation, estimate around 100k gas as a safe estimate
+            gasEstimate = 100000;
```
Luisfc68: why 100k?

Luisfc68: can't we estimate somehow? at least using gasleft() before and after the call

forge-scripts/tasks/RegisterPegin.s.sol
```diff
@@ -0,0 +1,505 @@
...
+        console.log("  BTC Network:", btcNetwork);
+
+        (bytes memory rawTx, bytes memory pmt, uint256 height) = fetchBtcTxData(
+            txId,
+            btcNetwork
+        );
+
+        // Estimate gas
+        console.log("\nEstimating gas...");
+        uint256 gasStart = gasleft();
```
Luisfc68, line 238: This is what I was talking about in the refundUserPegout comment

forge-test/tasks/PauseSystem.t.sol
```diff
@@ -0,0 +1,206 @@
...
+            "COLLATERAL_MANAGEMENT_ADDRESS",
+            vm.toString(address(collateral))
+        );
+    }
+
+    function test_CheckStatus() public view {
+        console.log("\n=== TEST CHECK STATUS ===\n");
+
+        // Call the actual script's checkStatus function
+        pauseScript.checkStatus();
```
Luisfc68, line 83: shouldn't you assert that the contracts are paused here?

forge-test/Pause.t.sol
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    function test_AllowsViewFunctionsToContinueWorkingWhenPaused() public view {
+        assertTrue(flyoverDiscovery.getProvidersId() >= 0);
+        assertEq(collateralManagement.getMinCollateral(), TEST_MIN_COLLATERAL);
+        assertTrue(pegInContract.getMinPegIn() > 0);
+        assertTrue(pegOutContract.dustThreshold() > 0);
+    }
+
+    function test_AllowsNonPausableFunctionsToContinueWorking() public pure {
```
Luisfc68: this test is not implemented

forge-test/deployment/ChangeOwnerToMultiSig.t.sol
```diff
@@ -0,0 +1,158 @@
...
+import {LiquidityBridgeContract} from "../../contracts/legacy/LiquidityBridgeContract.sol";
+import {LiquidityBridgeContractProxy} from "../../contracts/legacy/LiquidityBridgeContractProxy.sol";
+import {LiquidityBridgeContractAdmin} from "../../contracts/legacy/LiquidityBridgeContractAdmin.sol";
+
+/**
+ * @title ChangeOwnerToMultiSigTest
+ * @notice Test for the ChangeOwnerToMultiSig deployment script
+ * @dev Tests ownership transfer pattern to multisig
+ */
+contract ChangeOwnerToMultiSigTest is Test {
```
Luisfc68, line 17: I think the original tests validates that the new owner can perform only owner operations

forge-test/discovery/Registration.t.sol
```diff
@@ -0,0 +1,223 @@
...
+        address lp2 = makeAddr("lp2");
+        address lp3 = makeAddr("lp3");
+
+        vm.deal(lp1, 100 ether);
+        vm.deal(lp2, 100 ether);
+        vm.deal(lp3, 100 ether);
+
+        // Register LP1
+        vm.prank(lp1);
+        vm.expectEmit(false, false, false, false);
```
Luisfc68: I think is ok if we assert the event for the 3 providers

forge-test/pegin/CallForUser.t.sol
Luisfc68, line 1: I think the following tests are missing:
- sending the pegin to a smart contract
- revert if balance isn't enough
- test unsuccessful call (the call fails but the cfu doesn't)
- revert if the gas limit of cfu is not enough
- reentrancy test

forge-test/pegin/DerivationAddress.t.sol
```diff
@@ -0,0 +1,233 @@
...
+            cmInitData
+        );
+        collateralManagement = CollateralManagementContract(
+            payable(address(cmProxy))
+        );
+    }
+
+    // ============ validatePegInDepositAddress function tests ============
+
+    function test_ValidatePegInDepositAddress_FunctionExists() public {
```
Luisfc68: I think is ok if you want to include this in the integration tests but it should have at least a minimum validation in unitary level using hardcoded values

forge-test/pegin/RegisterPegIn.t.sol
```diff
@@ -0,0 +1,996 @@
...
+        bridgeMock.setHeader(
+            HEIGHT_MOCK + uint256(quote.depositConfirmations) - 1,
+            nConfHeader
+        );
+
+        // Don't call callForUser - test without it
+
+        // Register should revert due to insufficient amount
+        vm.prank(fullLp);
+        vm.expectRevert(); // AmountTooLow from Quotes
```
Luisfc68: can you add the assertion of the error type in this one?

forge-test/pegin/RegisterPegIn.t.sol
```diff
@@ -0,0 +1,996 @@
...
+            HEIGHT_MOCK + uint256(quote.depositConfirmations) - 1,
+            nConfHeader
+        );
+
+        // Don't call callForUser (LP didn't deliver)
+        uint256 userBalanceBefore = user.balance;
+
+        // Register by someone else (not LP) - LP gets penalized
+        vm.prank(registerCaller);
+        pegInContract.registerPegIn(
```
Luisfc68, line 595: This test is not asserting the events that should occur in this scenario

forge-test/pegin/RegisterPegIn.t.sol
```diff
@@ -0,0 +1,996 @@
...
+        );
+
+        // Advance time to after call deadline
+        vm.warp(quote.agreementTimestamp + quote.callTime + 1);
+
+        // LP calls callForUser late
+        vm.prank(fullLp);
+        pegInContract.callForUser{value: quote.value}(quote);
+
+        // Register by someone else - should penalize LP
```
Luisfc68, line 644: the penalize event is not being checked

forge-test/pegin/RegisterPegIn.t.sol
```diff
@@ -0,0 +1,996 @@
...
+        bridgeMock.setPegin{value: peginAmount}(quoteHash);
+        bridgeMock.setHeader(HEIGHT_MOCK, firstHeader);
+        bridgeMock.setHeader(
+            HEIGHT_MOCK + uint256(quote.depositConfirmations) - 1,
+            nConfHeader
+        );
+
+        // Register should revert
+        vm.prank(fullLp);
+        vm.expectRevert(); // AmountTooLow from Quotes
```
Luisfc68, line 690: can we assert the AmountTooLow error?

forge-test/pegin/RegisterPegIn.t.sol
```diff
@@ -0,0 +1,996 @@
...
+
+        // Verify LP got the full amount (including failed change)
+        assertEq(
+            pegInContract.getBalance(fullLp),
+            lpBalanceBefore + peginAmount + extraPaid - quote.productFeeAmount,
+            "LP should receive all funds when change payment fails"
+        );
+    }
+
+    // Note: Reentrancy tests would require deploying malicious contracts that attempt
```
Luisfc68: Can you explain why is not possible to deploy the malicious contracts?

forge-test/pegin/RegisterPegIn.t.sol
```diff
@@ -0,0 +1,996 @@
...
+        bridgeMock.setPegin{value: peginAmount}(quoteHash);
+        bridgeMock.setHeader(HEIGHT_MOCK, firstHeader);
+        bridgeMock.setHeader(
+            HEIGHT_MOCK + uint256(quote.depositConfirmations) - 1,
+            nConfHeader
+        );
+
+        // Call for user - will fail because wallet rejects
+        vm.prank(fullLp);
+        pegInContract.callForUser{value: quote.value}(quote);
```
Luisfc68, line 870: in this test asserting the event is particularly important, because the last argument tells wether the call was successful or not

forge-test/pegin/Withdraw.t.sol
Luisfc68, line 1: the reentrancy test is not included

forge-test/pegout/Deposit.t.sol
```diff
@@ -0,0 +1,427 @@
...
+            abi.encodeWithSelector(
+                Flyover.InsufficientAmount.selector,
+                sentAmount,
+                totalVal
+            )
+        );
+        pegOutContract.depositPegOut{value: sentAmount}(quote, signature);
+    }
+
+    function test_DepositPegOut_RevertsIfQuoteIsExpiredByDate() public {
```
Luisfc68: this scenario can happen either if deposit date limit or expire date have passed and I think this tests only validates one of them

Luisfc68: what I mean is that by expiring both at the same time the tests may pass if the revert condition is not properly implemented (e.g. reverting only if deposit date limit is expired)

forge-test/pegout/Deposit.t.sol
```diff
@@ -0,0 +1,427 @@
...
+            )
+        );
+        pegOutContract.depositPegOut{value: getTotalValue(quote)}(
+            quote,
+            wrongSignature
+        );
+    }
+
+    function test_DepositPegOut_RevertsIfQuoteAlreadyCompleted() public {
+        // Note: Testing quote completion requires full refundPegOut flow with BTC transactions
```
Luisfc68: The validation of the full cycle can be done in here, the btc data doesn't need to be accurate, can be mocks as in the original hardhat implementation

forge-test/pegout/Hashing.t.sol
```diff
@@ -0,0 +1,177 @@
...
+                Flyover.IncorrectContract.selector,
+                address(pegOutContract),
+                wrongContract
+            )
+        );
+        pegOutContract.hashPegOutQuote(quote);
+    }
+
+    function test_HashPegOutQuote_HashesPegOutQuoteProperly() public view {
+        // Note: Like PegIn hashing tests, we verify determinism rather than exact hashes
```
Luisfc68, line 53: I think this approach makes sense, the only problem I see is that is isn't sensitive to modifications in the hash function, for example, let's say we add a new field, this approach will pass regardless if the hash function includes the new field in the hash or not, in the case of the specific hash, if we add a new field the test will fail, which is ok

forge-test/pegout/LpRefund.t.sol
```diff
@@ -0,0 +1,617 @@
...
+/// Full refundPegOut testing requires complex BTC infrastructure:
+/// - BTC transaction generation with proper scripts (P2PKH, P2SH, P2WPKH, P2WSH, P2TR)
+/// - Merkle proof creation and validation
+/// - Block header mocking with proper timestamps
+/// - Testing 5 address types × 10 amount precisions = 50+ combinations
+/// - SAT/WEI conversion and truncation logic
+/// - Penalization based on timing (transfer windows, block/time expiry)
+///
+/// These tests cover the main validation paths. Full BTC transaction testing
+/// with all address types and amounts is in the TypeScript integration tests.
```
Luisfc68: The blocks and merkle proofs can be mocks, if fact, in the original tests they are mocks too, the only "real" data we were using are the transactions, which can be included here, I think this file should include all the scenarios that the original version had

forge-test/pegout/LpRefund.t.sol
```diff
@@ -0,0 +1,617 @@
...
+            PARTIAL_MERKLE_TREE,
+            merkleHashes
+        );
+    }
+
+    function test_RefundPegOut_RevertsIfNullDataMalformed() public {
+        Quotes.PegOutQuote memory quote = createAndDepositQuote();
+        bytes32 quoteHash = pegOutContract.hashPegOutQuote(quote);
+
+        // Test that a malformed Bitcoin transaction (too short) reverts
```
Luisfc68, line 142: In this case I would rename the test, this particular name refers a cases where the content of the null script is not the one we expect

forge-test/pegout/LpRefund.t.sol
```diff
@@ -0,0 +1,617 @@
...
+            uint32(quote.expireDate + 1)
+        );
+        bridgeMock.setHeaderByHash(BLOCK_HEADER_HASH, header);
+        bridgeMock.setConfirmations(
+            int256(uint256(quote.transferConfirmations))
+        );
+
+        bytes memory btcTx = generateBtcTx(quote, quoteHash);
+
+        // Refund should succeed but emit penalization
```
Luisfc68, line 372: the test is not asserting the penalization event

forge-test/pegout/LpRefund.t.sol
```diff
@@ -0,0 +1,617 @@
...
+        vm.prank(pegOutLp);
+        vm.expectEmit(true, false, false, true);
```
Luisfc68, line 418: the test is not asserting the penalization event

forge-test/pegout/LpRefund.t.sol
```diff
@@ -0,0 +1,617 @@
...
+        pegOutContract.refundPegOut(
+            quoteHash,
+            btcTx,
+            BLOCK_HEADER_HASH,
+            PARTIAL_MERKLE_TREE,
+            merkleHashes
+        );
+    }
+
+    // Note: The TypeScript test suite includes 100+ additional parameterized tests:
```
Luisfc68: can't you inject data using a custom script as you did in the tasks translations?

forge-test/legacy/PegIn.t.sol
```diff
@@ -0,0 +1,1464 @@
...
+        partialMerkleTree = hex"02327049330a25d4d17e53e79f478cbb79c53a509679b1d8a1505c5697afb426";
+        merkleBranchHashes = new bytes32[](1);
+        merkleBranchHashes[
+            0
+        ] = 0x02327049330a25d4d17e53e79f478cbb79c53a509679b1d8a1505c5697afb326;
+    }
+
+    // ============ Tests ============
+
+    function test_CallContractForUser() public {
```
Luisfc68, line 329: this test is not asserting the events

Luisfc68 (changes requested):

forge-test/pegout/LpRefund.t.sol
Hakob23: added for all missing event emissions and error messages mentioned in the conversation

forge-test/pegout/LpRefund.t.sol
Hakob23: yes, the issue is resolved

forge-test/tasks/HashQuote.t.sol
Luisfc68: The script still doesn't have any assertions

forge-test/tasks/HashQuote.t.sol
Luisfc68: In this case is also pending

forge-test/deployment/ChangeOwnerToMultiSig.t.sol
Luisfc68: Still pending, you could just try to setProviderStatus with the new owner

forge-test/pegin/RegisterPegIn.t.sol
Luisfc68: still needs to verify the events

forge-test/pegin/RegisterPegIn.t.sol
Luisfc68: still pending

forge-test/pegin/RegisterPegIn.t.sol
Luisfc68: still pending

forge-test/pegout/Hashing.t.sol
Luisfc68: Can't you use an approach similar to the one used in the hash quote task to address this issue?

forge-test/pegout/LpRefund.t.sol
```diff
@@ -0,0 +1,854 @@
...
+        pegOutContract.refundPegOut(
+            quoteHash,
+            hex"010203",
+            BLOCK_HEADER_HASH,
+            PARTIAL_MERKLE_TREE,
+            merkleHashes
+        );
+    }
+
+    function test_RefundPegOut_RevertsIfNullDataScriptHasWrongSize() public {
```
Luisfc68, line 155: This test is correct and I'd leave it, but my comment was about the case where the null data script is correct in format but the quote hash it represents doesn't match the quote being refunded

Luisfc68: So a copy paste of this test with a diff OP_RETURN should work

Luisfc68 (commented):
The PR is almost done, just need a few minor fixes

Luisfc68:
I'm merging the PR to unblock depending PRs. The fixes can be addressed in a new one as they're minor

PR #372, opened 2025-11-12
Fly 1130/pegout validation extension

Hakob23:
**What**

[The task](https://rsklabs.atlassian.net/jira/software/c/projects/FLY/boards/682?assignee=712020%3A06f65950-a686-4803-85cf-442c880e9bcf&selectedIssue=FLY-1130&useStoredSettings=true)

**Why**

The idea of this task is to add a new function to the `PegOutContract` which receives the transaction that is going to be used to call `refundPegOut `and validate if its ok


test/pegout/LpRefund.t.sol
```diff
@@ -702,6 +702,298 @@ contract LpRefundTest is PegOutTestBase {
...
+    }
+
+    function test_ValidatePegout_RevertsIfNullDataMalformed() public {
+        Quotes.PegOutQuote memory quote = createAndDepositQuote();
+        bytes32 quoteHash = pegOutContract.hashPegOutQuote(quote);
+
+        // Test that a malformed Bitcoin transaction (too short) reverts
+        // Using a minimal invalid tx hex"010203" instead of properly formed tx
+        vm.prank(pegOutLp);
+        vm.expectRevert(); // MalformedTransaction
```
Luisfc68: you're not asserting the error type

test/pegout/LpRefund.t.sol
```diff
@@ -702,6 +702,298 @@ contract LpRefundTest is PegOutTestBase {
...
+        bytes32 quoteHash = pegOutContract.hashPegOutQuote(quote);
+
+        // Generate BTC tx with WRONG address
+        Quotes.PegOutQuote memory wrongAddressQuote = quote;
+        wrongAddressQuote.depositAddress = new bytes(21); // Different address!
+        wrongAddressQuote.depositAddress[0] = 0x00; // Set version byte
+        bytes memory btcTx = generateBtcTx(wrongAddressQuote, quoteHash);
+
+        vm.prank(pegOutLp);
+        vm.expectRevert(); // InvalidDestination
```
Luisfc68, line 872: you're not asserting the error type

test/pegout/LpRefund.t.sol
```diff
@@ -702,6 +702,298 @@ contract LpRefundTest is PegOutTestBase {
...
+            btcTx
+        );
+
+        // Verify returned quote matches
+        assertEq(returnedQuote.value, quote.value);
+        assertEq(returnedQuote.lpRskAddress, quote.lpRskAddress);
+        assertEq(returnedQuote.lbcAddress, quote.lbcAddress);
+    }
+
+    function test_ValidatePegout_DoesNotRequireConfirmations() public {
```
Luisfc68, line 897: isn't this test the same as the one above?

test/pegout/LpRefund.t.sol
```diff
@@ -702,6 +702,298 @@ contract LpRefundTest is PegOutTestBase {
...
+
+        bytes memory btcTx = generateBtcTx(quote, quoteHash);
+
+        vm.prank(pegOutLp);
+        Quotes.PegOutQuote memory returnedQuote = pegOutContract.validatePegout(
+            quoteHash,
+            btcTx
+        );
+
+        // Verify all quote fields are correct
```
Luisfc68, line 927: These aren't all the fields, you can use the hash here but if you don't want to do that then I think you should add all the fields

test/pegout/LpRefund.t.sol
```diff
@@ -702,6 +702,298 @@ contract LpRefundTest is PegOutTestBase {
...
+        vm.expectRevert(
+            abi.encodeWithSelector(
+                IPegOut.QuoteAlreadyCompleted.selector,
+                quoteHash
+            )
+        );
+        pegOutContract.validatePegout(quoteHash, btcTx);
+    }
+
+    function test_ValidatePegout_AcceptsSameTransactionAsRefundPegOut() public {
```
Luisfc68, line 972: The idea of this test is great, but I think is implemented in the wrong way. You should run both validate and refund pegout using the same tx to ensure they respond consistently, now you're running them with 2 diff txs

Luisfc68 (changes requested):

Luisfc68 (approved):

PR #374, opened 2025-11-19
Fly 2081/update pegin pegout contracts to use access control

Hakob23:
**What**

[The task](https://rsklabs.atlassian.net/jira/software/c/projects/FLY/boards/682?assignee=712020%3A06f65950-a686-4803-85cf-442c880e9bcf&selectedIssue=FLY-2081)

**Why**

To decrease the amount of smart contracts and streamline the EmergencyPause mechanism


contracts/DaoContributor.sol
```diff
@@ -1,22 +1,19 @@
...
+/// @title AccessControlDaoContributorUpgradeable
 /// @notice This contract is used to handle the contributions to the DAO
 /// @author Rootstock Labs
 /// @dev Any contract that inherits from this contract will be able to collect DAO
 /// contributions according to the logic the child contract defines
-abstract contract OwnableDaoContributorUpgradeable is
+abstract contract AccessControlDaoContributorUpgradeable is
     ReentrancyGuardUpgradeable,
-    OwnableUpgradeable {
+    IAccessControl {
```
Luisfc68: Why are you implementing this interface? If you're using `_checkRole` to fix the dependency with the access control

Additionally I think you should receive in the initializer the admin role, as the admin might not be `0x00` if the default is changed

Luisfc68 (changes requested):

contracts/DaoContributor.sol
Hakob23: regarding `accessControl` it was a leftover from the previous setup. The admin is also initialized via initializer now

Luisfc68 (approved):

PR #418, opened 2026-02-11
feat: add fallback to pegout refunds

Luisfc68:
## What
Add balance/withdraw mechanism in pegout contract

## Why
To have a fallback mechanism in case refunds can't be paid

## Task
https://rsklabs.atlassian.net/browse/FLY-2204

### DIFF
--- a/src/PegOutContract.sol
+++ b/src/PegOutContract.sol
@@ -179,7 +179,8 @@ contract PegOutContract is
     }
 
     /// @inheritdoc IPegOut
-    function withdraw(address payable addr, uint256 amount) external nonReentrant override  {
+    function withdraw(address payable addr, uint256 amount) external nonReentrant override {
+        if (addr == address(0)) revert Flyover.InvalidAddress(addr);
         uint256 balance = _balances[msg.sender];
         if (balance < amount) {
             revert Flyover.NoBalance(amount, balance);
--- a/src/PegOutContract.sol
+++ b/src/PegOutContract.sol
@@ -43,6 +43,7 @@ contract PegOutContract is
 
     mapping(bytes32 => Quotes.PegOutQuote) private _pegOutQuotes;
     mapping(bytes32 => PegOutRecord) private _pegOutRegistry;
+    mapping(address => uint256) private _balances;
 
     /// @notice The dust threshold for the peg out. If the difference between the amount paid and the amount required
     /// is more than this value, the difference goes back to the user's wallet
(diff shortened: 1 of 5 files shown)


src/PegOutContract.sol
```diff
@@ -198,9 +215,9 @@ contract PegOutContract is
         }
 
         uint256 refundAmount = quote.value + quote.callFee + quote.gasFee;
-        (bool sent, bytes memory reason) = quote.lpRskAddress.call{value: refundAmount}("");
+        (bool sent,) = quote.lpRskAddress.call{value: refundAmount}("");
         if (!sent) {
```
AndresQuijano, line 224: What about triggering an event here (something like` emit RefundPaymentFailed(quoteHash, quote.lpRskAddress, refundAmount, reason); `so it's clear the operation failed.

src/PegOutContract.sol
```diff
@@ -221,9 +238,9 @@ contract PegOutContract is
         emit PegOutUserRefunded(quoteHash, addressToTransfer, valueToTransfer);
         _collateralManagement.slashPegOutCollateral(msg.sender, quote, quoteHash);
 
-        (bool sent, bytes memory reason) = addressToTransfer.call{value: valueToTransfer}("");
+        (bool sent,) = addressToTransfer.call{value: valueToTransfer}("");
         if (!sent) {
-            revert Flyover.PaymentFailed(addressToTransfer, valueToTransfer, reason);
+            _increaseBalance(addressToTransfer, valueToTransfer);
```
AndresQuijano, line 248: Same here: emit RefundPaymentFailed(quoteHash, quote.lpRskAddress, refundAmount, reason);

alexjavabraz (approved):

src/PegOutContract.sol
Luisfc68: I think you can infer that by checking the `BalanceIncrease` event. If later we add other scenario in this function where `BalanceIncrease` might be emitted we can add the other event, but I think right now only this one is enough and saves some gas

src/PegOutContract.sol
Luisfc68: I think you can infer that by checking the `BalanceIncrease` event. If later we add other scenario in this function where `BalanceIncrease` might be emitted we can add the other event, but I think right now only this one is enough and saves some gas

AndresQuijano (approved):

PR #424, opened 2026-02-16
Feature/fly 2184

Hakob23:
**What**

Timelock Control Integration

**Why**

Introduce a delay to the contract upgrades for increasing the reaction time in case of emergency during LPs resignation and withdrawal process

**Task**

https://rsklabs.atlassian.net/browse/FLY-2184


example.env
```diff
@@ -32,3 +32,17 @@ MULTISIG_OWNER_2_TESTNET=<owner2_address>
 MULTISIG_OWNER_3_TESTNET=<owner3_address>
 MULTISIG_OWNER_4_TESTNET=<owner4_address>
```
alexjavabraz, line 33: Will it be part of .env file? Can we add this encrypted and only decrypted in memory?

alexjavabraz (approved):

example.env
Hakob23: This is the template for .env file, not the actual .env file. In the real one, we have to replace with real values and it won't be seen or accesible outside of the local environment(let me know if I understood the comment correctly)

script/legacy/deployment/ChangeOwnerToTimelock.s.sol
```diff
@@ -0,0 +1,140 @@
...
+        vm.startBroadcast(deployerKey);
+
+        TimelockController timelock = new TimelockController(
+            cfg.timelockMinDelay,
+            proposers,
+            executors,
+            address(0)
+        );
+
+        _transferContractOwnership(proxyAddress, address(timelock));
```
Luisfc68: The timelock controller should be owner **only** of the proxy admin, because we might need to perform operations instantly in the contract (e.g. an emergency pause)

script/legacy/deployment/ChangeOwnerToTimelock.s.sol
```diff
@@ -0,0 +1,140 @@
...
+
+        address[] memory proposers = new address[](1);
+        proposers[0] = cfg.timelockProposer;
+
+        address[] memory executors = new address[](1);
+        executors[0] = cfg.timelockExecutor;
+
+        vm.startBroadcast(deployerKey);
+
+        TimelockController timelock = new TimelockController(
```
Luisfc68: perhaps we should allow to configure multiple proposers/acceptors in case is needed, not only one. Same for the admin, we should allow the script to receive one, even if probably in production it will be disabled

script/legacy/deployment/ChangeOwnerToTimelock.s.sol
Luisfc68, line 1: we should have this script also for the split architecture, because this is going to be merged for the v2.6.0 when that architecture is supposed to be already deployed

Luisfc68 (changes requested):

Luisfc68:
@Hakob23 we updated the v2.6.0 branch with the v2.5.0 fixes, pls update the feature branch accordingly when you push the next changes

script/legacy/deployment/ChangeOwnerToTimelock.s.sol
Hakob23: This is making the scripts slightly more complex, but yes, we can do it

Luisfc68 (approved):

script/deployment/ChangeOwnerToTimelock.s.sol
```diff
@@ -0,0 +1,153 @@
...
+
+        proposers = vm.parseJsonAddressArray(
+            json,
+            string.concat(".", networkKey, ".proposers")
+        );
+        executors = vm.parseJsonAddressArray(
+            json,
+            string.concat(".", networkKey, ".executors")
+        );
+
```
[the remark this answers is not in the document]
Hakob23: The issues is already address by adding regtest fields in the addresses

script/legacy/deployment/ChangeOwnerToTimelock.s.sol
```diff
@@ -0,0 +1,183 @@
...
+
+        proposers = vm.parseJsonAddressArray(
+            json,
+            string.concat(".", networkKey, ".proposers")
+        );
+        executors = vm.parseJsonAddressArray(
+            json,
+            string.concat(".", networkKey, ".executors")
+        );
+
```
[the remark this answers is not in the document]
Hakob23: the same here

PR #427, opened 2026-02-16
Feature/fly 2212

Hakob23:
**What**

Adding support for EIP-712 standard contracts for signature verification

**Why**

To avoid EOA bypass of signature verification via constructor call and also not introducing back tx.origin from legacy contract as it also can be manipulated.

**Task**

https://rsklabs.atlassian.net/browse/FLY-2212


src/libraries/SignatureValidator.sol
```diff
@@ -1,32 +1,73 @@
...
 
-import { ECDSA } from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import {IERC1271} from "@openzeppelin/contracts/interfaces/IERC1271.sol";
+import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 
 library SignatureValidator {
 
     using ECDSA for bytes32;
+    bytes4 internal constant EIP1271_MAGIC_VALUE_BYTES32 = IERC1271.isValidSignature.selector;
+    bytes4 internal constant EIP1271_MAGIC_VALUE_BYTES = 0x20c13b0b;
```
Luisfc68: this magic value doesn't match with the one of the specification https://eips.ethereum.org/EIPS/eip-1271

src/libraries/SignatureValidator.sol
```diff
@@ -1,32 +1,73 @@
...
 
     error IncorrectSignature(address expectedAddress, bytes32 usedHash, bytes signature);
     error ZeroAddress();
     /**
-        @dev Verfies signature against address
+        @dev Verifies signature against address.
+        @dev For EOAs, validates ECDSA over the EIP-191 prefixed hash.
```
Luisfc68: Now we're supporting eip 712, pls update this comment (and sync with the v2.5.0-fixes branch so you have those changes here)

src/FlyoverDiscovery.sol
```diff
@@ -188,7 +188,7 @@ contract FlyoverDiscovery is
         address providerAddress,
         uint256 collateralAmount
     ) private view {
-        if (providerAddress != msg.sender || providerAddress.code.length != 0) revert NotEOA(providerAddress);
+        if (providerAddress != msg.sender) revert NotAuthorized(msg.sender);
```
Luisfc68: In my opinion, I think is ok if we want to extend the signature validator to be capable of validating contract signatures, but I'm a bit worried about allowing contracts to be registered as liquidity providers, I think it might be a good call to keep this check EOA only so we can make the proper analysis of the impact of this decision, if we conclude is fine to allow smart contracts, then the signature validator is already working and we just need to update this check. Anyway I'll send this question to security as well

Luisfc68 (changes requested):

Luisfc68 (approved):

PR #431, opened 2026-02-17
Fix/fly 2216

Hakob23:
**What**

Fixing comparison condition with `quote.value`

**Why**

The `_registerCallNotDone` function in `PegInContract.sol` contains an arithmetic underflow that causes a guaranteed revert when `quote.value` is 0. By making `if (quote.callOnRegister && refundAmount > quote.value - 1)` comparison non-strict (`if (quote.callOnRegister && refundAmount >= quote.value )` ) this issue is resolved.

**Task**

https://rsklabs.atlassian.net/browse/FLY-2216


Luisfc68:
This PR has the wrong target branch

Luisfc68:
Closing this as is a duplicate of #433

PR #442, opened 2026-03-03
Adding Invariant Tests

Hakob23:
**What**

Adding invariant tests, including both isolated and integrated test suites. 
The changes include
- addition of an isolated test suite for `discoveryContract`
- addition of `SystemInvariant` test suite for integrated invariant testing
- addition of handler contracts for all invariant tests
- refactoring of invariant tests to use the handlers

**Task**

https://rsklabs.atlassian.net/browse/FLY-2167

### DIFF
--- a/src/PegInContract.sol
+++ b/src/PegInContract.sol
@@ -7,6 +7,7 @@ import {
 import {EIP712Upgradeable} from "@openzeppelin/contracts-upgradeable/utils/cryptography/EIP712Upgradeable.sol";
 import {ReentrancyGuardUpgradeable} from "@openzeppelin/contracts-upgradeable/utils/ReentrancyGuardUpgradeable.sol";
 import {BtcUtils} from "@rsksmart/btc-transaction-solidity-helper/contracts/BtcUtils.sol";
+import {OpCodes} from "@rsksmart/btc-transaction-solidity-helper/contracts/OpCodes.sol";
 import {EmergencyPause} from "./EmergencyPause/EmergencyPause.sol";
 import {IBridge} from "./interfaces/IBridge.sol";
 import {ICollateralManagement, CollateralManagementSet} from "./interfaces/ICollateralManagement.sol";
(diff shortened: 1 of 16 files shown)


src/PegInContract.sol
Luisfc68, line 1: update this PR with the latest changes so you can use the library

test/invariant/PegInInvariant.t.sol
```diff
@@ -1,120 +1,96 @@
...
-        if (balance == 0) return;
-
-        amount = bound(amount, 1, balance);
+    // ============ Invariant Tests ============
 
-        vm.prank(lp);
-        try pegInContract.withdraw(amount) {
-            ghost_totalWithdrawn += amount;
-        } catch {}
+    /// @notice Contract balance must cover sum of all LP internal balances
```
Luisfc68: This invariant makes sense but I think it doesn't always apply. Imagine this scenario:
1. User pegs in
2. LP doesn't behave
3. User registers pegin manually (or anyone who isn't the LP)
4. The refund fails for X reason
5. Now `quote.rskRefundAddress` has balance and its not an LP

Luisfc68: Honestly I'm not sure what is better, to update this invariant to reflect a different situation, or to update the set up to simulate what I described above. I'll leave that up to you

test/invariant/PegOutInvariant.t.sol
```diff
@@ -1,77 +1,154 @@
...
+            FuzzSelector({addr: address(handler), selectors: selectors})
         );
     }
 
     // ============ Invariant Tests ============
 
-    /// @notice Contract balance should never underflow
-    function invariant_NoUnderflow() public view {
-        uint256 contractBalance = address(pegOutContract).balance;
+    /// @notice Contract balance must equal deposited minus refunded minus withdrawn
```
Luisfc68, line 52: what happens if for example the user deposits more than the total value of the quote but the diff is less than the `dustThreshold`? In that case the balance of the contract would be > 0 because the refunds and withdraws wouldn't consider that diff

test/invariant/SystemInvariant.t.sol
```diff
@@ -0,0 +1,407 @@
...
+
+        assertGe(
+            balance,
+            totalCollateral + punisherRewards + userRewards + penalties,
+            "INVARIANT VIOLATED: Collateral contract insolvent"
+        );
+    }
+
+    /// @notice PegInContract balance must cover sum of all LP balances
+    function invariant_PegInSolvency() public view {
```
Luisfc68, line 118: question: why is this considered a system invariant and not a pegin contract invariant?

Luisfc68 (approved):
The PR is approved, my comments are more related to the definition of some invariants, but not to the code itself, I think the discussion about the invariants should be taken separately as its not so simple, but as PoC the code LGTM

test/invariant/PegInInvariant.t.sol
Hakob23: Imo, adding another invariant for this scenario is better

test/invariant/PegOutInvariant.t.sol
Hakob23: yes, we should also take into account the overpay

test/invariant/SystemInvariant.t.sol
Hakob23: Because the invariant is checked while providers are registered/resigned/status-toggled and other contracts are mutating and the address set used in the invariant comes from system handler’s provider registry, not a PegIn-only static LP list:

PR #444, opened 2026-03-12
Feature/fly 2232

Hakob23:
**What**

Implement Hard and Soft Pause functionality, and add corresponding test cases

**The Task**

https://rsklabs.atlassian.net/browse/FLY-2232?atlOrigin=eyJpIjoiZDUwNzE3ODY3OTUxNDk5NmFhMDFmOWEwOTg5MmMwY2EiLCJwIjoiaiJ9

### DIFF
--- a/src/PauseRegistry.sol
+++ b/src/PauseRegistry.sol
@@ -40,8 +40,6 @@ contract PauseRegistry is
 
     event EmergencyPaused(address indexed by, string reason);
     event EmergencyUnpaused(address indexed by);
-    error InvalidPauseLevel(uint8 level);
-    error AlreadyPaused();
 
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor() {
@@ -57,29 +55,29 @@ contract PauseRegistry is
     }
 
     /// @inheritdoc IPauseRegistry
-    function pause(string calldata reason) external onlyRole(PAUSER_ROLE) {
-        PauseRegistryStorage storage $ = _getPauseRegistryStorage();
-        if ($.pauseLevel != IPauseRegistry.PauseLevel.None) {
-            revert AlreadyPaused();
-        }
-        _setPauseLevel(IPauseRegistry.PauseLevel.Soft);
-        $.pauseReason = reason;
-        emit EmergencyPaused(msg.sender, reason);
-    }
-
-    /// @inheritdoc IPauseRegistry
-    function unpause() external onlyRole(PAUSER_ROLE) {
-        _setPauseLevel(IPauseRegistry.PauseLevel.None);
-        PauseRegistryStorage storage $ = _getPauseRegistryStorage();
-        $.pauseReason = "";
-        emit EmergencyUnpaused(msg.sender);
-    }
-
-    /// @inheritdoc IPauseRegistry
     function setPauseLevel(
-        IPauseRegistry.PauseLevel level
+        IPauseRegistry.PauseLevel level,
+        string calldata reason
     ) external onlyRole(PAUSER_ROLE) {
+        _setPauseLevelWithReason(level, reason);
+    }
+
+    function _setPauseLevelWithReason(
+        IPauseRegistry.PauseLevel level,
+        string memory reason
+    ) internal {
+        PauseRegistryStorage storage $ = _getPauseRegistryStorage();
+        bool wasPaused = $.pauseLevel != IPauseRegistry.PauseLevel.None;
+        if (level != IPauseRegistry.PauseLevel.None) {
+            $.pauseReason = reason;
+        }
         _setPauseLevel(level);
+        bool isPaused = level != IPauseRegistry.PauseLevel.None;
+        if (!wasPaused && isPaused) {
+            emit EmergencyPaused(msg.sender, $.pauseReason);
+        } else if (wasPaused && !isPaused) {
+            emit EmergencyUnpaused(msg.sender);
+        }
     }
 
     /// @inheritdoc IPauseRegistry
@@ -192,6 +190,7 @@ contract PauseRegistry is
             level == IPauseRegistry.PauseLevel.None
         ) {
             $.pauseTimestamp = 0;
+            $.pauseReason = "";
         }
 
         if (
--- a/src/PauseRegistry.sol
+++ b/src/PauseRegistry.sol
@@ -10,8 +10,8 @@ import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Ini
 import {IPauseRegistry} from "./interfaces/IPauseRegistry.sol";
 
 /// @title PauseRegistry
-/// @notice Centralized registry for pause state; all Flyover contracts read from here
-/// @dev Only accounts with PAUSER_ROLE can pause/unpause. Uses namespaced storage.
+/// @notice Centralized registry for pause state; all Flyover contracts read from here.
+/// @dev Level 0 = normal, 1 = soft (no new business), 2 = hard (full freeze). Uses namespaced storage.
 contract PauseRegistry is
     Initializable,
     AccessControlDefaultAdminRulesUpgradeable,
@@ -20,8 +20,17 @@ contract PauseRegistry is
     /// @custom:storage-location erc7201:rsk.flyover.PauseRegistry
     struct PauseRegistryStorage {
         bool paused;
+        uint8 pauseLevel;
         uint64 pauseTimestamp;
         string pauseReason;
+        HardPause[] hardPauses;
+    }
+
+    struct HardPause {
+        uint64 startTimestamp;
+        uint64 endTimestamp; // 0 while ongoing
+        uint64 startBlock;
+        uint64 endBlock; // 0 while ongoing
     }
 
     bytes32 public constant PAUSER_ROLE = keccak256("PAUSER_ROLE");
@@ -31,6 +40,7 @@ contract PauseRegistry is
 
     event EmergencyPaused(address indexed by, string reason);
     event EmergencyUnpaused(address indexed by);
+    error InvalidPauseLevel(uint8 level);
 
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor() {
@@ -47,27 +57,30 @@ contract PauseRegistry is
 
     /// @inheritdoc IPauseRegistry
     function pause(string calldata reason) external onlyRole(PAUSER_ROLE) {
+        _setPauseLevel(1);
         PauseRegistryStorage storage $ = _getPauseRegistryStorage();
-        if ($.paused) return;
-        $.paused = true;
-        $.pauseReason = reason;
         $.pauseTimestamp = uint64(block.timestamp);
         emit EmergencyPaused(msg.sender, reason);
     }
 
     /// @inheritdoc IPauseRegistry
     function unpause() external onlyRole(PAUSER_ROLE) {
+        _setPauseLevel(0);
         PauseRegistryStorage storage $ = _getPauseRegistryStorage();
-        if (!$.paused) return;
-        $.paused = false;
-        $.pauseReason = "";
         $.pauseTimestamp = 0;
         emit EmergencyUnpaused(msg.sender);
     }
 
     /// @inheritdoc IPauseRegistry
+    function setPauseLevel(uint8 level) external onlyRole(PAUSER_ROLE) {
+        if (level > 2) revert InvalidPauseLevel(level);
+        _setPauseLevel(level);
+    }
+
+    /// @inheritdoc IPauseRegistry
     function paused() external view returns (bool) {
-        return _getPauseRegistryStorage().paused;
+        PauseRegistryStorage storage $ = _getPauseRegistryStorage();
+        return $.pauseLevel != 0;
     }
 
     /// @inheritdoc IPauseRegistry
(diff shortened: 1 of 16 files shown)


src/PauseRegistry.sol
```diff
@@ -20,8 +20,17 @@ contract PauseRegistry is
     /// @custom:storage-location erc7201:rsk.flyover.PauseRegistry
     struct PauseRegistryStorage {
         bool paused;
+        uint8 pauseLevel;
         uint64 pauseTimestamp;
         string pauseReason;
+        HardPause[] hardPauses;
```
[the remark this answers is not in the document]
Hakob23: We do not have any deployed version yet, so the current packing is not a problem. Moreover, the first 3 fields together take only 1 32 byte storage. If we move it down, then we will occupy an extra 32 byte space

src/PauseRegistry.sol
```diff
@@ -47,27 +57,30 @@ contract PauseRegistry is
 
     /// @inheritdoc IPauseRegistry
     function pause(string calldata reason) external onlyRole(PAUSER_ROLE) {
+        _setPauseLevel(1);
         PauseRegistryStorage storage $ = _getPauseRegistryStorage();
-        if ($.paused) return;
-        $.paused = true;
-        $.pauseReason = reason;
         $.pauseTimestamp = uint64(block.timestamp);
         emit EmergencyPaused(msg.sender, reason);
```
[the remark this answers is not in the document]
Hakob23: The suggestion is correct but not complete:  `pauseStatus` should also be updated. Will add the whole suggestion as a manual fix

src/PauseRegistry.sol
```diff
@@ -47,27 +57,30 @@ contract PauseRegistry is
...
-        $.paused = false;
-        $.pauseReason = "";
         $.pauseTimestamp = 0;
         emit EmergencyUnpaused(msg.sender);
     }
 
+    /// @inheritdoc IPauseRegistry
+    function setPauseLevel(uint8 level) external onlyRole(PAUSER_ROLE) {
+        if (level > 2) revert InvalidPauseLevel(level);
+        _setPauseLevel(level);
```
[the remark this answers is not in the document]
Hakob23: The catch is correct, but there is a cleaner approach to track the pause timestamp

src/PegOutContract.sol
```diff
@@ -232,12 +235,20 @@ contract PegOutContract is
...
-    function refundUserPegOut(bytes32 quoteHash) external nonReentrant override {
+    function refundUserPegOut(bytes32 quoteHash) external nonReentrant whenNotHardPaused override {
         Quotes.PegOutQuote memory quote = _pegOutQuotes[quoteHash];
 
         if (quote.lbcAddress == address(0)) revert Flyover.QuoteNotFound(quoteHash);
+
+        uint256 depositTs = _pegOutRegistry[quoteHash].depositTimestamp;
+        uint256 depositBlock = _pegOutRegistry[quoteHash].depositBlock;
+        uint256 pauseOverlap = pauseRegistry().computePauseOverlap(depositTs, block.timestamp);
+        uint256 pauseBlocks = pauseRegistry().computePauseOverlapBlocks(depositBlock, block.number);
```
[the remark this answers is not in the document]
Hakob23: While practically we avoid this issue, as during the upgrade no pending quotes will exist, it is still better to have this hedge for consistency with legacy quote types. A general solution implementing both suggestions(the one below referencing to the same issue) will be committed manually

src/PegOutContract.sol
```diff
@@ -370,13 +381,18 @@ contract PegOutContract is
...
         if (firstConfirmationTimestamp > expectedConfirmationTime) {
             return true;
         }
 
-        // penalize if LP is refunding after expiration
-        if (block.timestamp > quote.expireDate || block.number > quote.expireBlock) {
+        uint256 depositTs = _pegOutRegistry[quoteHash].depositTimestamp;
+        uint256 depositBlock = _pegOutRegistry[quoteHash].depositBlock;
+        uint256 pauseOverlap = pauseRegistry().computePauseOverlap(depositTs, block.timestamp);
+        uint256 pauseBlocks = pauseRegistry().computePauseOverlapBlocks(depositBlock, block.number);
```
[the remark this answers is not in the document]
Hakob23: Answered above

src/PauseRegistry.sol
```diff
@@ -47,27 +57,30 @@ contract PauseRegistry is
 
     /// @inheritdoc IPauseRegistry
     function pause(string calldata reason) external onlyRole(PAUSER_ROLE) {
+        _setPauseLevel(1);
         PauseRegistryStorage storage $ = _getPauseRegistryStorage();
-        if ($.paused) return;
-        $.paused = true;
-        $.pauseReason = reason;
         $.pauseTimestamp = uint64(block.timestamp);
```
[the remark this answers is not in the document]
Hakob23: Instead of escalating pause level, its better to have OZ-style behavior of reverting the `pause()` function when already paused, as an unintentional call could escalate the pause level and cause unwanted behavior. The escalation logic is possible only through `setPauseLevel()`. The changes are committed manually

src/PegInContract.sol
```diff
@@ -596,8 +591,19 @@ contract PegInContract is
             nConfirmationsHeader
         );
 
-        // penalize if the call was not made on time
-        if (callTimestamp > nConfirmationsTimestamp + quote.callTime) {
+        uint256 pauseOverlap = pauseRegistry().computePauseOverlap(
+            quote.agreementTimestamp,
```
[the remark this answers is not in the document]
Hakob23: committed with corresponding test suite

src/interfaces/IPauseRegistry.sol
```diff
@@ -2,14 +2,18 @@
...
     function pause(string calldata reason) external;
 
-    /// @notice Unpauses the system
+    /// @notice Unpauses the system — sets level to 0
     function unpause() external;
 
-    /// @notice Returns whether the system is paused
+    /// @notice Set pause level (PAUSER_ROLE). On enter 2 a HardPause entry is appended; on leave 2 it is closed.
+    /// @param level 0, 1, or 2
+    function setPauseLevel(uint8 level) external;
```
Luisfc68: what is the difference between this and `pause`? is the same doing `setPauseLevel(1)` and `pause('reason')`? If yes maybe the interface is a bit confusing and could be improved

Luisfc68: same with unpause

Luisfc68: in my point of view, we could either have `pause()|unpause()|hardPause()` or `setPauseLevel()` but having both might be too confusing

src/interfaces/IPauseRegistry.sol
```diff
@@ -21,4 +25,33 @@ interface IPauseRegistry {
...
         view
         returns (bool isPaused, string memory reason, uint64 since);
+
+    /// @notice Current pause level: 0 = normal, 1 = soft (no new business), 2 = hard (full freeze)
+    function pauseLevel() external view returns (uint8);
+
+    /// @notice Number of hard-pause log entries (for reverse iteration)
+    function hardPausesCount() external view returns (uint256);
+
+    /// @notice One hard-pause log entry (start/end timestamp and block; end 0 while ongoing)
```
Luisfc68, line 38: question: the reason why this uses index instead of returning the list is to delegate the responsibility of building the list to the caller by using `hardPausesCount` + this right?

src/PauseRegistry.sol
```diff
@@ -77,7 +92,119 @@ contract PauseRegistry is
...
+        uint8 prev = $.pauseLevel;
+        $.pauseLevel = level;
+        $.paused = (level != 0);
+        if (prev == 0 && level != 0) {
+            $.pauseTimestamp = uint64(block.timestamp);
+        } else if (prev != 0 && level == 0) {
+            $.pauseTimestamp = 0;
+        }
+
+        if (prev == 2 && level != 2) {
```
Luisfc68: I think we should enforce the state transitions to be more strict e.g.
only allow 
0 -> 1 
1 -> 0
2 -> 0
0 -> 2
anything else should return an error, also if you try to make 1 -> and is already paused should return an error (maybe the PR already does that, haven't got to that part yet)

Luisfc68: that will also make this code easier

src/PauseRegistry.sol
```diff
@@ -77,7 +92,119 @@ contract PauseRegistry is
...
+            uint256 effectiveEnd = endBlock;
+            uint256 pEndOrNow = pEndBlock == 0 ? block.number : pEndBlock;
+            if (pEndOrNow < effectiveEnd) effectiveEnd = pEndOrNow;
+            if (effectiveEnd > effectiveStart) {
+                totalPauseBlocks += effectiveEnd - effectiveStart;
+            }
+        }
+    }
+
+    function _setPauseLevel(uint8 level) internal {
```
Luisfc68: I suggest to handle the hard and soft pauses separately in two private functions, that will make the code easier to review and less error prone to maintain

src/PauseRegistry.sol
```diff
@@ -77,7 +92,119 @@ contract PauseRegistry is
...
+        if (prev == 0 && level != 0) {
+            $.pauseTimestamp = uint64(block.timestamp);
+        } else if (prev != 0 && level == 0) {
+            $.pauseTimestamp = 0;
+        }
+
+        if (prev == 2 && level != 2) {
+            HardPause[] storage pauses = $.hardPauses;
+            uint256 len = pauses.length;
+            if (len > 0) {
```
Luisfc68, line 202: in which cases this array would be empty if the system was previously paused?

src/PauseRegistry.sol
Luisfc68, line 1: I recommend to add an enum for the pause levels, that will use the same 0,1,2 values but will make the code more readable

src/PauseRegistry.sol
```diff
@@ -77,7 +92,119 @@ contract PauseRegistry is
...
+            }
+            HardPause storage p = pauses[i];
+            uint64 pEnd = p.endTimestamp;
+            if (pEnd != 0 && !(pEnd > startTimestamp)) break;
+            uint256 effectiveStart = startTimestamp;
+            if (p.startTimestamp > effectiveStart) effectiveStart = p.startTimestamp;
+            uint256 effectiveEnd = endTimestamp;
+            uint256 pEndOrNow = pEnd == 0 ? block.timestamp : pEnd;
+            if (pEndOrNow < effectiveEnd) effectiveEnd = pEndOrNow;
+            if (effectiveEnd > effectiveStart) {
```
Luisfc68, line 146: if which case this might be false? isn't it better to revert in that case instead of ignoring it?

src/PauseRegistry.sol
```diff
@@ -77,7 +92,119 @@ contract PauseRegistry is
...
+            uint64 startBlock,
+            uint64 endBlock
+        )
+    {
+        HardPause storage p = _getPauseRegistryStorage().hardPauses[index];
+        return (p.startTimestamp, p.endTimestamp, p.startBlock, p.endBlock);
+    }
+
+    /// @inheritdoc IPauseRegistry
+    function computePauseOverlap(uint256 startTimestamp, uint256 endTimestamp)
```
Luisfc68, line 127: maybe we should revert if not start < end

src/PauseRegistry.sol
```diff
@@ -77,7 +92,119 @@ contract PauseRegistry is
...
+            uint256 pEndOrNow = pEnd == 0 ? block.timestamp : pEnd;
+            if (pEndOrNow < effectiveEnd) effectiveEnd = pEndOrNow;
+            if (effectiveEnd > effectiveStart) {
+                totalPauseTime += effectiveEnd - effectiveStart;
+            }
+        }
+    }
+
+    /// @inheritdoc IPauseRegistry
+    function computePauseOverlapBlocks(uint256 startBlock, uint256 endBlock)
```
Luisfc68, line 153: same comments as in `computePauseOverlap`

alexjavabraz (approved):
LGTM

Luisfc68 (changes requested):

src/interfaces/IPauseRegistry.sol
Hakob23: `setPauseLevel()` is actually more universal version for `pause()` function. Imo `pause()` also could be removed, as I don't see a reason to keep it when we have 2 functions doing the same thing, however, both functions were mentioned in the requirements, on my understanding to keep some sort of compatibility with OZ Pausable interface.

src/interfaces/IPauseRegistry.sol
Hakob23: exactly

src/PauseRegistry.sol
Hakob23: Why should we put that restriction? Couldn't there be a scenario, when we want to deescalate the pause?

src/PauseRegistry.sol
Hakob23: hardPauses tracks only level 2 (hard) pauses, not all pauses, so the array can be empty even if the system was previously paused in softer phases

src/PauseRegistry.sol
Hakob23: Yes, we can also just use words (NONE, SOFT, HARD) if we are using enums anyway

src/PauseRegistry.sol
Hakob23: when effectiveEnd equals or less than effectiveStart. Otherwise we would receive the underflow error. But here I don't think that revert is necessary

src/PauseRegistry.sol
Hakob23: I don't think so. But we can discuss it

src/interfaces/IPauseRegistry.sol
Luisfc68: I see, I agree about `setPauseLevel()`. Perhaps the task was not clear enough that for this version we just need `setPauseLevel()`. We can remove `pause()|unpause()` then

src/PauseRegistry.sol
Luisfc68: Maybe, my comment was because if we reduce the space of valid transitions, it is less error prone and therefore has less possibilities we introduce a vulnerability. Just the scenario of pausing has low changes of happening (or at least its the objective) so the scenario where we pause and escalate or deescalate has even lower. Thats why I consider a valid idea reducing the behavior

src/PauseRegistry.sol
Luisfc68: closing after discussion in slack

src/PauseRegistry.sol
Luisfc68: closing after discussion in slack

Luisfc68 (approved):

src/interfaces/IPauseRegistry.sol
Hakob23: I have removed `pause()|unpause()` as discussed, and added `string reason` parameter to `setPauseLevel()`

PR #461, opened 2026-05-13
Fix/FLY-2299 - Limit provider data

Luisfc68:
## What
Add limits to provider metadata

## Why
To prevent abuse of the storage that might derivate later in expensive views if called from another contract

## Task
https://rsklabs.atlassian.net/browse/FLY-2299

### DIFF
--- a/src/FlyoverDiscovery.sol
+++ b/src/FlyoverDiscovery.sol
@@ -246,6 +246,10 @@ contract FlyoverDiscovery is
         return _liquidityProviders[providerId];
     }
 
+    /// @notice Validates provider metadata fields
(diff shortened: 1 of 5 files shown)


Hakob23 (approved):
LGTM!

src/FlyoverDiscovery.sol
```diff
@@ -22,6 +22,8 @@ contract FlyoverDiscovery is
 {
     /// @notice The version of the contract
     string constant public VERSION = "1.0.0";
+    uint256 constant private _MAX_PROVIDER_NAME_LENGTH = 256;
```
VaWheel, line 25: I don't think provider name can be that big. Can we use 50?

PR #466, opened 2026-05-18
feat: add gas snapshot comparison tools and workflows

Hakob23:
## What
Standardize local gas-diff behavior and keep the comparator implementation in TypeScript.

## Why
Local gas comparison flow had inconsistencies and lint/pre-commit blockers.  
This change makes the local command path predictable and hook-safe while staying aligned with existing CI intent.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change 
- [ ] Documentation update
- [x] Refactoring (no functional changes, no api changes)
- [x] Performance improvement
- [x] Test updates
- [ ] Security fix
- [ ] Deployment/Infrastructure changes

## Affected part of the project
- [ ] Management UI / API
- [ ] PegIn flow
- [ ] PegOut flow
- [x] Utility scripts
- [x] Configuration files
- [ ] Metrics and alerting

## Related Issues
[Jira ticket](https://rsklabs.atlassian.net/browse/FLY-2303)

## How to test

1. create a new branch
2. `make snapshot`
3. commit changes
4. `make gas-compare`
5. Verify `gas_comparison.csv` is generated and command output is as expected.

The .yml illustration will be shown automatically after newly pushed branches

## Additional Notes
Comparator remains in TypeScript by design.


script/compare_gas.ts
```diff
@@ -0,0 +1,55 @@
...
+      const gas = Number.parseInt(match[2], 10);
+      gasValues[testName] = gas;
+    }
+  });
+
+  return gasValues;
+}
+
+const previousCommit = execSync("git rev-parse HEAD~1").toString().trim();
+
```
TravellerOnTheRun, line 70: `compare_gas.ts` crashes with an opaque git error when `HEAD~1` does not contain `.gas-snapshot` (for example right after the first snapshot commit).

```suggestion
const previousCommit = execSync("git rev-parse HEAD~1").toString().trim();

const currentGasSnapshotContent = readFileSync(".gas-snapshot", "utf-8");
let previousGasSnapshotContent: string;
try {
  previousGasSnapshotContent = execSync(
    `git show ${previousCommit}:.gas-snapshot`,
    { encoding: "utf-8" }
  ).trim();
} catch {
  console.error(
    `No .gas-snapshot at ${previousCommit}. Commit a baseline first or use a ref that contains the file.`
  );
  process.exit(1);
}
```

TravellerOnTheRun (changes requested):
## Blocker — security or funds risk

None.

## Required before merge

- **`gas-snapshot-compare` documents `--base` and `--base-file`, but `compare_gas.ts` always compares against `HEAD~1` and does not parse arguments.** — The Makefile advertises baseline options that the script does not implement (AC-1 predictable local path).

- **`compare_gas.ts` crashes with an opaque git error when `HEAD~1` does not contain `.gas-snapshot` (for example right after the first snapshot commit).** — see inline suggestion on `script/compare_gas.ts` ~L26

## Recommended

- **The gas diff workflow runs on `pull_request` and `push` to `main`, while this PR merges into `QA-Test` — consider whether `QA-Test` should be included for push-triggered runs.** — Post-merge pushes to `QA-Test` will not trigger the workflow as written.

- **`gas-compare` and `gas-snapshot-compare` both call the same script with different documented behavior; consider consolidating or implementing the differentiated behavior.** — Two targets, one code path; only one advertises alternate baselines today.

- **The gas CSV only lists tests present in the current snapshot, so removed or renamed tests do not show up as regressions in the report.** — Rows are built only from current snapshot keys.

## Optional nit

- **The sticky comment step checks for `pull_request_target`, but the workflow only lists `pull_request` — consider simplifying the condition to avoid confusion.** — Harmless for this PR’s `pull_request` runs; reduces maintainer confusion.

- **Ensure `gas_comparison.csv` is listed in `.gitignore` so local compare output is never committed by mistake.** — Script writes a generated CSV at repo root.

## Summary-only findings

### `gas-snapshot-compare` documents `--base` and `--base-file`, but `compare_gas.ts` always compares against `HEAD~1` and does not parse arguments.

The Makefile `gas-snapshot-compare` target documents optional `--base=<ref>` and `--base-file=<path>` and passes them to `script/compare_gas.ts`. The script ignores all CLI input and always resolves the previous snapshot from `HEAD~1` via `git show`. Authors following documented `make gas-snapshot-compare` options will not get the behavior described in help text or the PR test steps.

**Suggested direction:** Either wire `process.argv` for `--base` / `--base-file` (and test both paths), or simplify the Makefile target and docs to match the `HEAD~1`-only behavior until flags are implemented.

### The gas diff workflow runs on `pull_request` and `push` to `main`, while this PR merges into `QA-Test` — consider whether `QA-Test` should be included for push-triggered runs.

After merge to `QA-Test`, pushes to that branch will not run the gas diff workflow because `on.push.branches` is limited to `main`. If `QA-Test` is where integration happens before `main`, post-merge gas reporting on direct pushes may be absent. This PR still gets PR-scoped runs via `pull_request`.

### `gas-compare` and `gas-snapshot-compare` both call the same script with different documented behavior; consider consolidating or implementing the differentiated behavior.

`gas-compare` and `gas-snapshot-compare` both invoke the same TypeScript entrypoint today. Only the latter advertises alternate baselines (currently non-functional). Consider one documented target or distinct implementations.

### The sticky comment step checks for `pull_request_target`, but the workflow only lists `pull_request` — consider simplifying the condition to avoid confusion.

The sticky comment step’s `if` includes `pull_request_target`, which is not declared under `on:`. Harmless for normal `pull_request` runs but confusing for maintainers.

### The gas CSV only lists tests present in the current snapshot, so removed or renamed tests do not show up as regressions in the report.

When a test disappears from `.gas-snapshot`, it never appears in `gas_comparison.csv` because rows are built only from the current snapshot keys. Reviewers watching gas drift may miss removals unless they diff snapshots manually.

### Ensure `gas_comparison.csv` is listed in `.gitignore` so local compare output is never committed by mistake.

`compare_gas.ts` writes `gas_comparison.csv` in the repo root. Confirm `.gitignore` excludes it so it is not committed alongside the committed `.gas-snapshot` baseline.

TravellerOnTheRun (dismissed):
LGTM

Freshenext (approved):

PR #469, opened 2026-05-20
refactor: add foundry upgrades

Luisfc68:
## What
Deploy using foundry upgrades 

## Why
Because is safer to upgrade the contracts in the future

## Task
https://rsklabs.atlassian.net/browse/FLY-2325


script/deployment/DeployLibraries.s.sol
Hakob23, line 1: Does this upgrade make deploying the libraries instead of using already deployed ones necessary or we just want to use the newer version of libs each time we do an upgrade?

Hakob23 (approved):
Apart from this [comment](https://github.com/rsksmart/liquidity-bridge-contract/pull/469/changes#r3283076539) that simply needs clarification, LGTM

script/deployment/DeployLibraries.s.sol
Luisfc68: OZ Upgrades treats external libraries as already deployed dependencies of the implementation bytecode, instead of a part of the implementation deployment process (like new XXX() does), we can still reuse libraries, but we need to make sure they are deployed separately, unlike the current version where if they aren't the same forge script deploys them for you

Luisfc68: so if you go and try to deploy without having the addresses of the deployed libraries the OZ plugin will fail in one of the safety validations

PR #451, opened 2026-04-14
Test/fly 2243

Hakob23:
**What**

Integrate [Halmos](https://github.com/a16z/halmos) as the formal proof verification tool for the Flyover contracts, with PoC symbolic proofs for CollateralManagement.

**Why**

Per security requirements, formal proof verification is the remaining testing layer to add. This PR integrates the tool and validates it with initial proofs; formal specification scope will be handled in a follow-up discussion.

**Why Halmos (and not other tools)**

- Native Foundry fit: The project is Foundry-first; Halmos uses Solidity tests and integrates directly with existing workflows.
- Easier integration: No new spec DSL required (unlike Certora/CVL); tests can written as check_* proofs in Solidity.

- Open-source and free: Better fit for team iteration.
- Toolchain fit: Works with current Python + Foundry setup.

- Certora is stronger in terms of automatized workflow for cross-contract flow testing but requires a new language(CVL) and a commercial licensing. Halmos can provide the same coverage with more manual set-up. So the trade-off of not including a new language in the stack is worth it imo
- Kontrol is stronger in terms of providing unbounded loop verifications, but has a much more complex setup. Additionally, the core logic doesn't have loop-specific code blocks, so using Kontrol just for unbounded loop verifications would be an overkill. 

**Type of Change**

-  New feature (non-breaking change)

-  Build/CI/Tooling

**Test Plan**

`make test-formal`

**Deployment & Ops Impact**

-  No deployment impact
- Local prerequisite for developers running proofs: Halmos installed (`pip install halmos` )

**Jira Ticket**

https://rsklabs.atlassian.net/browse/FLY-2243

**Reviewer Notes**

- Formal tests run through `make test-formal`
- Halmos profile is configured to ensure compatibility in local environments
- Scope is intentionally PoC-level; broader formal specs/invariants will be defined in a follow-up phase

PR #474, opened 2026-05-27
chore: add Python virtual environment setup script and update Makefil…

Hakob23:
### What

Add Python virtual environment setup script and update Makefile and documentation

### Notes

- To be merged after  https://github.com/rsksmart/liquidity-bridge-contract/pull/451
- Addressing https://github.com/rsksmart/liquidity-bridge-contract/pull/451#discussion_r3306253831


scripts/setup-python.sh
```diff
@@ -0,0 +1,81 @@
...
+#      system doesn't have one.
+#   2. Otherwise, look for python3.12 or python3.11 on PATH (Halmos requires
+#      Python >= 3.11).
+#   3. If neither is available, fail with installation instructions.
+
+set -euo pipefail
+
+VENV_DIR=".venv"
+TARGET_PYTHON="3.12"   # used by uv; matches CI's actions/setup-python step
+REQ_DEV="requirements-dev.txt"
```
Luisfc68: do we really need separated requirement files?

Hakob23: The reason is that we require different versions for Halmos and Slither workflow containers. But if we remove Python installation from `npm prepare` hook and manually call `make python-setup` then we could get rid of the file. What option do you lean to more?

Luisfc68: Ok, I don't see a problem on removing the hook, or converting the hook into make pyhon-setup

Luisfc68 (approved):

PR #509, opened 2026-07-16
Feat/fly 2459

Hakob23:
## What

Implements `FlyoverConfigurations`, the single on-chain source of peg-in fee,
confirmation tiers, and amount limits that every party (SDK, LPS, settlement)
reads. Behind the S0-frozen `IFlyoverConfigurations` interface, with admin
changes gated by a two-step time lock and re-validated against immutable
deployment bounds. Ships provisional regtest values annotated with the
stakeholder decision that owns each real value.

New files:
- `src/FlyoverConfigurations.sol` — implements the frozen peg-in interface
  (fee calc, confirmation tiers, amount limits, time-locked admin changes),
  ERC-7201 storage, proxied + initializable per repo pattern.
- `src/libraries/FlyoverConfigurationsRegtest.sol` — provisional regtest seed
  config, immutable bounds, and time-lock delay (consumed later by S5 deploy).

## Why

The commit-first redesign removes the off-chain quote, so the fee, confirmation
tiers, and limits that used to live in the quote need one owner-set on-chain
source every party trusts equally. The fixed-fee floor is a *security*
parameter (decision 2·D): if it drops below worst-case RSK gas during
congestion, an attacker can make min-amount peg-ins no LP will serve, and every
LP eats the global slash. Immutable bounds ensure an admin mistake (or a
compromised role) cannot set absurd values.

## Type of Change

- [x] New feature (non-breaking change)
- [x] Security hardening

## Affected Areas

- [x] PegIn flow (`registerPegIn`, `callForUser`, quote validation)
      (new config contract read by the peg-in flow; no existing peg-in code changed yet)

## Risk & Security Review

- [x] Access control and authorization paths reviewed (queue/apply gated by `DEFAULT_ADMIN_ROLE`)
- [x] Reentrancy and external call paths reviewed (no external calls; `receive()` reverts)
- [x] Storage layout / upgrade safety reviewed (ERC-7201 namespaced; see below)
- [x] Time/block/confirmation assumptions reviewed (time lock uses `block.timestamp`; tiers keyed by amount)
- [ ] BTC tx output/index assumptions reviewed (if pegout validation touched) — N/A
- [x] No secrets or sensitive values added

### Storage layout (ERC-7201)

Namespace `rsk.flyover.FlyoverConfigurations`
slot `0x13aa2a37a5354fe7c5dcced2a6c33933ec66091f98f22792660cd2862f158700`:

| Field       | Type             | Notes                                  |
|-------------|------------------|----------------------------------------|
[...]

PR #510, opened 2026-07-16
Feat/fly 2460

Hakob23:
## What

Adds the Foundry unit-test suite for `FlyoverConfigurations` under
`test/configurations/**`, pinning every number the protocol will read from the
contract: fee math, confirmation-tier lookups, amount limits, all getters, the
admin authorization path, and the two-step time-locked change flow. Contract
under test ships in a separate PR (S2.1 / FLY-2459); this is the sibling test
suite (S2.2 / FLY-2460).

New files (own `*TestBase.sol` per repo convention):
- `test/configurations/ConfigurationsTestBase.sol` — ERC1967-proxy deploy, seed
  config, explicit deployment bounds, tier/config helpers.
- `Fee.t.sol` — floor, satoshi rounding, hand-computed values.
- `Tiers.t.sol` — boundaries, above-last-tier, single-tier config.
- `Getters.t.sol` — configuration, bounds, pending change, delay, version.
- `AdminGating.t.sol` — role gating + every config-validation revert.
- `Timelock.t.sol` — all four timelock cases + exact-eta, no-queue, events.
- `Scaffolding.t.sol` — init-once, admin role, seed-validated-at-init,
  `receive()` rejection, ERC-7201 slot placement.

## Why

The commit-first redesign removes the off-chain quote, so the fee, confirmation
tiers, and limits become on-chain values every party (SDK, LPS, settlement)
trusts equally. Before any consumer reads them, each value and each revert path
must be pinned by a test — especially the fixed-fee floor, a security parameter
(decision 2·D), and the immutable bounds that guard even a compromised admin
role.

## Type of Change

- [x] Tests only

## Affected Areas

- [x] PegIn flow (config values consumed by the peg-in flow; no production code changed)
- [x] CI workflows (`ci`, `fuzz`, `formal`, `slither`, `codeql`) — suite runs under `make test-unit`

## Risk & Security Review

- [x] Access control and authorization paths reviewed (queue/apply gated by `DEFAULT_ADMIN_ROLE`; non-admin reverts covered)
- [x] Reentrancy and external call paths reviewed (config is read-mostly; `receive()` rejection covered)
- [x] Storage layout / upgrade safety reviewed (ERC-7201 base-slot placement asserted via `vm.load`)
- [x] Time/block/confirmation assumptions reviewed (delay, exact-eta, before-delay all covered)
- [ ] BTC tx output/index assumptions reviewed — N/A

## Test Plan

- [ ] `npm run lint:sol` — N/A (globs `src/**` only; test files are not linted)
- [ ] `npm run lint:ts`
- [x] `npm test` / `make test-unit` — 604 passed, 0 failed (includes the 47 new configuration tests)

Case coverage (Spec):
- Fee math: zero-amount floor, floor at min amount, mid amount, large-amount
  no-overflow, satoshi-rounding parity with `Quotes.checkAgreedAmount`, all
  asserted against hand-computed values.
- Tier lookup: per-tier, exact boundary (inclusive), just-above, above the last
  tier, single-tier config.
- Limits + getters: configuration, bounds, pending (empty/queued/cleared),
  delay, version.
- Admin gating: role required on queue and apply; empty / non-ascending /
  equal-adjacent tiers; each field out-of-bounds; `InvalidPercentageFee`;
  `InvalidAmountLimits`.
- Timelock, all four cases: (1) queue→apply after delay; (2) apply-before-delay
  reverts; (3) out-of-bounds reverts at both queue and apply time (apply-time
  guard exercised via a stored-pending corruption); (4) second queue replaces
  pending. Plus exact-eta, `NoQueuedChange`, `ChangeQueued`/`ChangeApplied`.

## Deployment & Ops Impact

- [x] No deployment impact (test-only)

## Related Issues

https://rsklabs.atlassian.net/browse/FLY-2460

## Reviewer Notes

- Suite lives in `test/configurations/**`, which `make test-unit` includes
  automatically (it is not under the excluded `test/{fuzz,invariant,differential,formal}` paths).
- `ConfigurationsTestBase` uses an ERC1967 proxy (not Transparent) so `owner`
  can call admin functions directly; a TransparentUpgradeableProxy would route
  the admin's calls to the proxy admin instead of the implementation.
- Deliberate choice worth a look: the `percentageFee` upper *bound* in the test
  base is set above the 100% denominator (`20_000`) specifically so the
  `InvalidPercentageFee` guard is reachable and provably independent of the
  bound check — otherwise that revert path is shadowed by the bound.
- Case 3's apply-time bound revert is proven by queuing a valid change and then
  `vm.store`-corrupting the stored pending value before applying, since
  `queueChange` itself blocks out-of-bounds values through the normal path.
  
  
## Note

To be merged **AFTER**  https://github.com/rsksmart/liquidity-bridge-contract/pull/509

### DIFF
--- a/test/configurations/Scaffolding.t.sol
+++ b/test/configurations/Scaffolding.t.sol
@@ -17,7 +17,14 @@ contract ScaffoldingTest is ConfigurationsTestBase {
 
     function test_initializeOnce_reverts() public {
         vm.expectRevert(); // Initializable: InvalidInitialization
-        config.initialize(owner, ADMIN_DELAY, TIMELOCK_DELAY, _seedConfig(), _boundsMin(), _boundsMax());
+        config.initialize(
+            owner,
+            ADMIN_DELAY,
+            TIMELOCK_DELAY,
+            _seedConfig(),
+            _boundsMin(),
+            _boundsMax()
+        );
     }
 
     function test_adminRoleAssigned() public view {
@@ -37,7 +44,14 @@ contract ScaffoldingTest is ConfigurationsTestBase {
 
         bytes memory initData = abi.encodeCall(
             FlyoverConfigurations.initialize,
-            (owner, ADMIN_DELAY, TIMELOCK_DELAY, badSeed, _boundsMin(), _boundsMax())
+            (
+                owner,
+                ADMIN_DELAY,
+                TIMELOCK_DELAY,
+                badSeed,
+                _boundsMin(),
+                _boundsMax()
+            )
         );
 
         vm.expectRevert(
--- a/test/configurations/Scaffolding.t.sol
+++ b/test/configurations/Scaffolding.t.sol
@@ -16,7 +16,7 @@ contract ScaffoldingTest is ConfigurationsTestBase {
     }
 
     function test_initializeOnce_reverts() public {
-        vm.expectRevert(); // Initializable: InvalidInitialization
+        vm.expectRevert(abi.encodeWithSignature("InvalidInitialization()"));
         config.initialize(
             owner,
             ADMIN_DELAY,
--- a/test/configurations/Scaffolding.t.sol
+++ b/test/configurations/Scaffolding.t.sol
@@ -1,0 +1,71 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.25;
+
+import {ConfigurationsTestBase} from "./ConfigurationsTestBase.sol";
+import {ERC1967Proxy} from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
+import {FlyoverConfigurations} from "../../src/FlyoverConfigurations.sol";
+import {IFlyoverConfigurations} from "../../src/interfaces/IFlyoverConfigurations.sol";
+import {Flyover} from "../../src/libraries/Flyover.sol";
+
+/// @title ScaffoldingTest
+/// @notice Proxy/initialize wiring: init-once, admin role assignment, seed validation at init,
+/// value rejection, and ERC-7201 storage namespace placement.
+contract ScaffoldingTest is ConfigurationsTestBase {
+    function setUp() public {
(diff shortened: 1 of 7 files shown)


test/configurations/Scaffolding.t.sol
```diff
@@ -0,0 +1,85 @@
...
+/// @title ScaffoldingTest
+/// @notice Proxy/initialize wiring: init-once, admin role assignment, seed validation at init,
+/// value rejection, and ERC-7201 storage namespace placement.
+contract ScaffoldingTest is ConfigurationsTestBase {
+    function setUp() public {
+        _deploy();
+    }
+
+    function test_initializeOnce_reverts() public {
+        vm.expectRevert(); // Initializable: InvalidInitialization
```
TravellerOnTheRun: Assert the exact reinitialization error instead of accepting any revert.

```suggestion
 vm.expectRevert(abi.encodeWithSignature("InvalidInitialization()"));
```

TravellerOnTheRun (changes requested):
## Blocker — security or funds risk

None.

## Required before merge

- **Add coverage for the public fee-percentage denominator getter.**
  Add a direct assertion that `config.FEE_PERCENTAGE_DENOMINATOR()` equals the expected denominator so the suite pins the value exposed by the contract rather than only its local mirror.
- **Assert the exact reinitialization error instead of accepting any revert.** — see inline suggestion on `test/configurations/Scaffolding.t.sol` ~L19
- **Verify every confirmation-tier value survives queue and apply.**
  Compare each queued tier's `maxAmount` and `confirmations` against both the pending and applied configurations, including the second tier, so the test proves the complete nested array survives the timelock flow.

## Recommended

None.

## Optional nit

None.

test/configurations/Scaffolding.t.sol
Hakob23: fixed

TravellerOnTheRun (approved):

PR #515, opened 2026-07-30
feat(FLY-2521): per-network zero-address BTC placeholders

Freshenext:
## Summary

- Both BTC placeholders in `PegInDerivation` now return the P2PKH **zero address** (`version ++ 20 zero bytes`) instead of a real regtest p2pkh, and they are **per network**: version byte `0x6f` on testnet/regtest, `0x00` on mainnet. `derivationValue` threads the `mainnet` flag to both accessors; `PegInAddressRegistry` sources it from the flag it already stores, and `_expectedDepositPkScript` gains the parameter so deposit-gating derives with the same flag as issuance.
- All derivation vectors re-pinned **per network**. A mainnet address is no longer a re-prefixed testnet address — the script hash itself differs — so the library and registry fixtures carry independent testnet/mainnet sets. Pins were computed offline by a separate implementation of the scheme rather than by calling the library under test, so a wrong version byte fails CI instead of being pinned as correct. The pitfall #1 (direct-tag) pin is unchanged, since that form never mixes the placeholders.
- `TODO(treasury)` added at both definitions: the bridge's failure-refund path pays this address, so with a zero hash those funds are **burned**. Accepted for the PoC; before production both must point at a monitored, recoverable, protocol-owned treasury address per network, and that swap rotates every derived address. NatSpec rewritten — constraints that still hold are kept (never empty, never the serving LP's address, identical at issuance and settlement); the claim that the value must be a real monitored address is dropped and now lives in the TODO.

## Not included: the end-to-end regtest proof

The ticket asks for a settled regtest peg-in with the zero placeholders, with the settle tx recorded here. **That proof is not in this PR, and this PR should not be treated as having satisfied it.**

`PegInContract.resolvePegIn` is a stub on `v3.0.0` — it reverts `ResolvePegInNotImplemented` — so there is no settlement path on this branch to settle anything through. The working settlement path and the harness that produced the reference settle tx `0x174ab0df…` live on the `dos-removal` branch, not here.

So the open question the ticket names — *does the bridge settle with a well-formed but zero-hash refund address, or reject it?* — remains **unanswered**. The library header records this explicitly as owed by the settlement work and as a gate before production, rather than as a property the library assumes.

## Verification

- `forge build` clean; `forge test` 752 passed. The 2 failures in `test/differential/DifferentialParity*.diff.t.sol` are pre-existing on clean `v3.0.0` (`setUp()` fails on a missing `.rskMainnet.LiquidityBridgeContract.address` entry in `addresses.json`) and are unrelated to this change.
- `prettier --check` clean; `solhint` reports the same 2 warnings before and after this change (no new lint).

### DIFF
--- a/src/libraries/PegInDerivation.sol
+++ b/src/libraries/PegInDerivation.sol
@@ -15,18 +15,18 @@ import {OpCodes} from "@rsksmart/btc-transaction-solidity-helper/contracts/OpCod
 /// validated end-to-end on a live regtest against the UNMODIFIED powpeg bridge (rskj 9.0.2,
 /// 2026-06-30, settle tx 0x174ab0df7cc86648a40d9b36da95fd4dacf2f18c135606141d532b7d1363c7de). That
 /// run used a real regtest p2pkh as the two BTC placeholders; they are now the per-network ZERO
-/// address (see {refundPlaceholderBtc}). The zero-address form is well-formed and non-empty, but no
-/// end-to-end run has yet settled with it — the settlement path does not exist on this branch
-/// (`PegInContract.resolvePegIn` reverts `ResolvePegInNotImplemented`). Proving the bridge accepts a
-/// zero-hash address is owed by the settlement work, and is a merge gate for production, not a
-/// property this library may assume:
+/// address (see {getRefundPlaceholderBtcAddress}). The zero-address form is well-formed and
+/// non-empty, but no end-to-end run has yet settled with it — the settlement path does not exist on
+/// this branch (`PegInContract.resolvePegIn` reverts `ResolvePegInNotImplemented`). Proving the
+/// bridge accepts a zero-hash address is owed by the settlement work, and is a merge gate for
+/// production, not a property this library may assume:
 ///
 ///   derivationArgumentsHash = keccak256(DERIVATION_DOMAIN ++ rskAddr)                    (step 1)
 ///   derivationValue         = keccak256(
 ///       derivationArgumentsHash
-///       ++ refundPlaceholderBtc(mainnet) // userRefundBtcAddress (21-byte per-network constant)
-///       ++ bytes20(pegInContract)        // lbcAddress = the contract that CALLS the bridge
-///       ++ lpPlaceholderBtc(mainnet)     // liquidityProviderBtcAddress (same per-network constant)
+///       ++ getRefundPlaceholderBtcAddress(isMainnet) // userRefundBtcAddress (21-byte constant)
+///       ++ bytes20(pegInContract)                    // lbcAddress = the contract CALLING the bridge
+///       ++ getLpPlaceholderBtcAddress(isMainnet)     // liquidityProviderBtcAddress (same constant)
 ///   )                                                                                    (step 2)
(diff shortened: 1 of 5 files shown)


src/libraries/PegInDerivation.sol
```diff
@@ -140,10 +177,10 @@ library PegInDerivation {
     /// string shown to the user. MUST stay a plain P2SH: the segwit-wrapped form is pitfall #2
     /// (-304).
     /// @param scriptHash Step 4's output
-    /// @param mainnet True for the mainnet version byte (0x05), false for testnet/regtest (0xC4)
+    /// @param isMainnet True for the mainnet version byte (0x05), false for testnet/regtest (0xC4)
     /// @return The 25-byte base58check payload of the deposit address
-    function depositAddressPayload(bytes20 scriptHash, bool mainnet) internal pure returns (bytes memory) {
-        bytes1 version = mainnet ? bytes1(0x05) : bytes1(0xC4);
+    function depositAddressPayload(bytes20 scriptHash, bool isMainnet) internal pure returns (bytes memory) {
+        bytes1 version = isMainnet ? bytes1(0x05) : bytes1(0xC4);
```
Luisfc68: You can move the version bytes to constants, similar to what you did with the addresses

test/libraries/PegInDerivation.t.sol
```diff
@@ -96,27 +132,61 @@ contract PegInDerivationTest is Test {
     }
 
     function test_RefundPlaceholderIsPinned() public pure {
```
Luisfc68, line 134: I think is worth to maintain the length assertion

Luisfc68: (I know there is a test below but it doesn't hurt)

Luisfc68 (commented):
None of the comments are blockers so feel free to address them or merge, up to you. 

One thing I think we need to reconsider is regarding the comments verbosity, I'm noticing that most of the recent PRs seem to be excesively commented (even when ignoring the parts that we're supposed to remove at the end like references to jira, the walkthrough, etc). Even to the point that adds noise to follow the code. If the intention is to keep this information in the repository, I'd suggest to move these to a `docs/` dir and focus the natspec only for the API related information, so the protocol and process description live in their own document, which is more organized and easier to share later (e.g. dev portal pipeline)

Luisfc68:
I noticed CI is failing, if you decide not to address the comments pls make sure to fix it before the merge

Luisfc68 (approved):
Approving as they're not blockers, but remember to fix the ci

test/libraries/PegInDerivation.t.sol
Freshenext: The length assertion is still there — it moved into the `_assertZeroPlaceholder` helper that this test calls, so it still runs on all four placeholder cases (refund/lp x testnet/mainnet):

https://github.com/rsksmart/liquidity-bridge-contract/blob/b89f0915/test/libraries/PegInDerivation.t.sol#L439-L443

The helper asserts length == 21, then the version byte, then that the remaining 20 bytes are zero, then the pin. Happy to inline it here as well if you'd still rather see it at the call site.

PR #516, opened 2026-07-31
feat(FLY-2575): derive the peg-in amount from the BTC deposit transaction

Freshenext:
Closes [FLY-2575](https://rsklabs.atlassian.net/browse/FLY-2575) (S4.3, under FLY-2476).

**Base**: `feature/FLY-2521` — this rebases on the per-network placeholder work, which threads the network flag through the derivation. Retarget to `v3.0.0` once FLY-2521 merges.

## The defect

`requestPegIn` received `uint256 amount` from the caller and nothing bound it to Bitcoin. The function was given `btcTxHash`, a hash, so it could not read the deposit's value even in principle. The only BTC-side check, `getBtcTransactionConfirmations`, proves a txid sits in a block at some depth — it says nothing about who the transaction paid or how much.

That unverified number decided the confirmation tier, the service fee, and the RBTC delivered to the user.

So the reachable attack was: take any confirmed Bitcoin txid, pair it with any registered `rskAddr`, declare a dust `amount`, front that dust as `msg.value`. Every check passed. The caller became the recorded claimer under `pegInId`, the user received dust, and every honest LP was locked out by `PegInAlreadyProcessed` — deliberately the first check in the function. No deposit to Flyover had to exist. Cost was dust plus gas; the result was a permanently unservable peg-in for a real user who really did deposit.

## The change

`requestPegIn` takes the raw witness-stripped deposit transaction in place of **both** `amount` and `btcTxHash`:

```solidity
function requestPegIn(
    address rskAddr,
    bytes calldata btcTxSerialized,
    bytes calldata opReturn,
    bytes32 btcBlockHash,
    uint256 merkleBranchPath,
    bytes32[] calldata merkleBranchHashes
) external payable returns (bytes32 pegInId);
```

The txid is hashed out of those same bytes with `BtcUtils.hashBtcTx`, so the SPV proof, the peg-in id and the amount provably describe one transaction — keeping the txid as an independent input would leave them free to disagree. The contract then derives the deposit `scriptPubKey` for `rskAddr` and reads the amount off the output paying it, scaled by `10 ** 10`.

The confirmation tier, the fee, the `msg.value` fronting equality, the claim record and the `PegInRequested` event all run against that derived amount with their current semantics and error types unchanged. Event field order and indexing are untouched.

Ordered checks in `requestPegIn`, with the new one at position 4:

1. `PegInAlreadyProcessed`
2. dependencies wired
3. `AddressNotRegistered`
4. **`DepositOutputNotFound(rskAddr, btcTxHash)`** ← new
5. `InsufficientConfirmations` (tier looked up with the derived amount)
6. `IncorrectFronting` (against the derived amount minus the derived fee)

## Shared helper

`PegInAddressRegistry` no longer holds its own output-matching implementation. `expectedDepositPkScript` and `matchedDepositValue` moved into `PegInDerivation`, which already declares itself the single source of truth for the derivation, and both contracts call them. `matchedDepositValue` returns a `(value, found)` pair rather than reverting, so each contract keeps its own named error — the registry's `DepositOutputNotFound(rskAddr)`, the peg-in contract's `DepositOutputNotFound(rskAddr, btcTxHash)`.

The registry's existing derivation and registration tests pass unchanged.

## The two decisions the ticket left open

**Where the helper lives** — `PegInDerivation`. The matching step is the consumer-side half of the same pipeline the library already owns, every step is already a separate function there, and the library's own doc comment states the byte-for-byte-agreement rule the matching has to honour.

**Split deposits** — **first match, not the sum.** Unchanged from the registry, so the value that gates registration and the value that fixes the peg-in amount stay identical; changing to a sum would have to change both call sites at once. First match understates a split deposit, and that direction is the safe one: the claimer fronts less than the bridge will release, so settlement covers the claim.

## Tests

31 tests across the migrated suites, all green; full unit run is 643 passing.

New coverage:

- A confirmed transaction unrelated to Flyover reverts `DepositOutputNotFound` — the case that closes the dust lockout.
- A real Flyover deposit derived for a *different* destination reverts `DepositOutputNotFound` (mismatched transaction).
- End to end on the original defect: the dust claim is rejected and the real depositor is served afterwards under the same `pegInId` the attacker tried to burn.
- A 50-RBTC deposit presented with only the low tier's confirmations reverts `InsufficientConfirmations`; the same depth still serves a deposit that really is small.
- Understating and overstating `msg.value` are separate `IncorrectFronting` cases, plus dust fronted against a real deposit.
- Satoshi-to-wei pinned to a literal fixture (123,456,789 sats == 1.23456789 RBTC), so an off-by-a-power-of-ten cannot pass CI.
- Atomicity: a failed deposit match writes no claim and delivers nothing.
- Ordering: `PegInAlreadyProcessed` still wins over the deposit match, since only the txid feeds the id.

Test fixtures now build real deposit transactions paying the address derived for their destination. A test that wants a bad claim has to build a bad transaction; it can no longer just pass a bad number.

## Notes for review

- **Deployment wiring.** `PegInContract` derives with `address(this)`, the live powpeg script, and its own `_mainnet` flag; the registry derives with its stored `pegInContract` and `isMainnet`. If a deployment sets those to disagree, every claim reverts `DepositOutputNotFound` — fail-closed, but a total outage. Neither `getPegInContract()` nor `getBridge()` is on the frozen `IPegInAddressRegistry`, so asserting agreement in `setPegInDependencies` needs a second frozen-interface amendment. Worth a follow-up ticket; not taken here.
- **Gas on the claim race.** The txid is hashed before the already-processed check, because the id is keyed on it. The loser of a race now pays one double-SHA256 over the transaction before reverting. Nothing else moved ahead of that check.
- **Not in scope.** `requestPegIn` still enforces neither `minAmount` nor `maxAmount` from `FlyoverConfigurations` (exceptions A4 and A7). Real, but it only becomes meaningful now that the amount is trustworthy, so it should land after this. No ticket exists yet.

## Cross-lane

- **LPS (FLY-2507)** — this is an ABI break. The LPS already holds the raw transaction for its settlement call, so it is a call-site change, not new plumbing. The contract side is asserted here; the caller update is asserted by FLY-2507. **Announce before merging.**
- **Spec** — the walkthrough step 11 amendment and the S0 frozen-signature amendment (S0.2, not yet created) belong to the doc owner and gate this merge. `IPegInCommitFirst`'s NatSpec is updated here to match.

### DIFF
--- a/src/PegInAddressRegistry.sol
+++ b/src/PegInAddressRegistry.sol
@@ -206,6 +206,10 @@ contract PegInAddressRegistry is
     }
 
     /// @notice Derives the on-chain P2SH scriptPubkey for a deposit output match.
+    /// @dev Reads the live powpeg script and hands the composition to
+    /// {PegInDerivation-expectedDepositPkScript}. The registry holds no derivation of its own:
+    /// `PegInContract.requestPegIn` matches deposits against that same helper, so the script that
+    /// gates registration is the script that fixes the peg-in amount.
     /// @param isMainnet Whether the derivation targets mainnet or testnet — the BTC placeholders
     /// mixed into the value are per-network, so this must match the flag used at issuance
     function _expectedDepositPkScript(address rskAddr, address pegInContract, IBridge bridge_, bool isMainnet)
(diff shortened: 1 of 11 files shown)


Freshenext:
Superseded: rebuilding on `v3.0.0` as the ticket specifies, rather than on the unmerged `feature/FLY-2521`. Replacement PR to follow.

PR #517, opened 2026-07-31
feat(FLY-2575): derive the peg-in amount from the BTC deposit transaction

Freshenext:
Closes [FLY-2575](https://rsklabs.atlassian.net/browse/FLY-2575) (S4.3, under FLY-2476).

**Base**: `v3.0.0`. One commit on top of `dbcf7b0`.

The ticket said to rebase on FLY-2521 first, but that branch is not merged and basing here on it would have pulled its two commits into this diff. Built on `v3.0.0` instead, so the derivation helper takes no network flag. When FLY-2521 lands, `expectedDepositPkScript` gains its `isMainnet` parameter and passes it through to `derivationValue` — a one-line change at each of the three call sites (registry, `PegInContract._derivePegInAmount`, and the test base). The ticket has been corrected.

## The defect

`requestPegIn` received `uint256 amount` from the caller and nothing bound it to Bitcoin. The function was given `btcTxHash`, a hash, so it could not read the deposit's value even in principle. The only BTC-side check, `getBtcTransactionConfirmations`, proves a txid sits in a block at some depth — it says nothing about who the transaction paid or how much.

That unverified number decided the confirmation tier, the service fee, and the RBTC delivered to the user.

So the reachable attack was: take any confirmed Bitcoin txid, pair it with any registered `rskAddr`, declare a dust `amount`, front that dust as `msg.value`. Every check passed. The caller became the recorded claimer under `pegInId`, the user received dust, and every honest LP was locked out by `PegInAlreadyProcessed` — deliberately the first check in the function. No deposit to Flyover had to exist. Cost was dust plus gas; the result was a permanently unservable peg-in for a real user who really did deposit.

## The change

`requestPegIn` takes the raw witness-stripped deposit transaction in place of **both** `amount` and `btcTxHash`:

```solidity
function requestPegIn(
    address rskAddr,
    bytes calldata btcTxSerialized,
    bytes calldata opReturn,
    bytes32 btcBlockHash,
    uint256 merkleBranchPath,
    bytes32[] calldata merkleBranchHashes
) external payable returns (bytes32 pegInId);
```

The txid is hashed out of those same bytes with `BtcUtils.hashBtcTx`, so the SPV proof, the peg-in id and the amount provably describe one transaction — keeping the txid as an independent input would leave them free to disagree. The contract then derives the deposit `scriptPubKey` for `rskAddr` and reads the amount off the output paying it, scaled by `10 ** 10`.

The confirmation tier, the fee, the `msg.value` fronting equality, the claim record and the `PegInRequested` event all run against that derived amount with their current semantics and error types unchanged. Event field order and indexing are untouched.

Ordered checks in `requestPegIn`, with the new one at position 4:

1. `PegInAlreadyProcessed`
2. dependencies wired
3. `AddressNotRegistered`
4. **`DepositOutputNotFound(rskAddr, btcTxHash)`** ← new
5. `InsufficientConfirmations` (tier looked up with the derived amount)
6. `IncorrectFronting` (against the derived amount minus the derived fee)

## Shared helper

`PegInAddressRegistry` no longer holds its own output-matching implementation. `expectedDepositPkScript` and `matchedDepositValue` moved into `PegInDerivation`, which already declares itself the single source of truth for the derivation, and both contracts call them. `matchedDepositValue` returns a `(value, found)` pair rather than reverting, so each contract keeps its own named error — the registry's `DepositOutputNotFound(rskAddr)`, the peg-in contract's `DepositOutputNotFound(rskAddr, btcTxHash)`.

The registry's existing derivation and registration tests pass unchanged.

## The two decisions the ticket left open

**Where the helper lives** — `PegInDerivation`. The matching step is the consumer-side half of the same pipeline the library already owns, every step is already a separate function there, and the library's own doc comment states the byte-for-byte-agreement rule the matching has to honour.

**Split deposits** — **first match, not the sum.** Unchanged from the registry, so the value that gates registration and the value that fixes the peg-in amount stay identical; changing to a sum would have to change both call sites at once. First match understates a split deposit, and that direction is the safe one: the claimer fronts less than the bridge will release, so settlement covers the claim.

## Tests

31 tests across the migrated suites, all green; full unit run is 636 passing.

New coverage:

- A confirmed transaction unrelated to Flyover reverts `DepositOutputNotFound` — the case that closes the dust lockout.
- A real Flyover deposit derived for a *different* destination reverts `DepositOutputNotFound` (mismatched transaction).
- End to end on the original defect: the dust claim is rejected and the real depositor is served afterwards under the same `pegInId` the attacker tried to burn.
- A 50-RBTC deposit presented with only the low tier's confirmations reverts `InsufficientConfirmations`; the same depth still serves a deposit that really is small.
- Understating and overstating `msg.value` are separate `IncorrectFronting` cases, plus dust fronted against a real deposit.
- Satoshi-to-wei pinned to a literal fixture (123,456,789 sats == 1.23456789 RBTC), so an off-by-a-power-of-ten cannot pass CI.
- Atomicity: a failed deposit match writes no claim and delivers nothing.
- Ordering: `PegInAlreadyProcessed` still wins over the deposit match, since only the txid feeds the id.

Test fixtures now build real deposit transactions paying the address derived for their destination. A test that wants a bad claim has to build a bad transaction; it can no longer just pass a bad number.

## Notes for review

- **Deployment wiring.** `PegInContract` derives with `address(this)` and the live powpeg script; the registry derives with its stored `pegInContract`. If a deployment sets those to disagree, every claim reverts `DepositOutputNotFound` — fail-closed, but a total outage. `getPegInContract()` is not on the frozen `IPegInAddressRegistry`, so asserting agreement in `setPegInDependencies` needs a second frozen-interface amendment. Worth a follow-up ticket; not taken here. FLY-2521 adds a network flag to the same derivation, which widens this surface by one more field.
- **Gas on the claim race.** The txid is hashed before the already-processed check, because the id is keyed on it. The loser of a race now pays one double-SHA256 over the transaction before reverting. Nothing else moved ahead of that check.
- **Not in scope.** `requestPegIn` still enforces neither `minAmount` nor `maxAmount` from `FlyoverConfigurations` (exceptions A4 and A7). Real, but it only becomes meaningful now that the amount is trustworthy, so it should land after this. No ticket exists yet.

## Cross-lane

- **LPS (FLY-2507)** — this is an ABI break. The LPS already holds the raw transaction for its settlement call, so it is a call-site change, not new plumbing. The contract side is asserted here; the caller update is asserted by FLY-2507. **Announce before merging.**
- **Spec** — the walkthrough step 11 amendment and the S0 frozen-signature amendment (S0.2, not yet created) belong to the doc owner and gate this merge. `IPegInCommitFirst`'s NatSpec is updated here to match.

### DIFF
--- a/src/FlyoverConfigurations.sol
+++ b/src/FlyoverConfigurations.sol
@@ -49,8 +49,14 @@ contract FlyoverConfigurations is
     uint256 public constant FEE_PERCENTAGE_DENOMINATOR = 10_000;
 
     /// @notice 1 satoshi expressed in wei; fees are rounded down to a satoshi boundary so on-chain
-    /// fees agree with the bridge. Mirrors `Quotes.SAT_TO_WEI_CONVERSION`.
-    uint256 public constant SAT_TO_WEI_CONVERSION = 10 ** 10;
+    /// fees agree with the bridge.
+    /// @dev The public home of the scale for the commit-first path: this getter is what the SDK
+    /// and every LPS read. The value itself is defined once in {Flyover} so the peg-in contract
+    /// can share it without an external call — see that declaration for why it cannot live here.
+    /// `Quotes.SAT_TO_WEI_CONVERSION` is the legacy quote path's own copy, left alone because that
+    /// path's ABI is frozen; `test/configurations/Fee.t.sol` asserts the two agree so they cannot
+    /// drift. `PegOutContract` holds a third, private copy.
+    uint256 public constant SAT_TO_WEI_CONVERSION = Flyover.SAT_TO_WEI_CONVERSION;
 
     // ERC-7201: keccak256(abi.encode(uint256(keccak256("rsk.flyover.FlyoverConfigurations")) - 1)) &
     // ~bytes32(uint256(0xff))
--- a/src/PegInContract.sol
+++ b/src/PegInContract.sol
@@ -453,8 +453,10 @@ contract PegInContract is
 
     /// @notice Reads the peg-in amount out of the deposit transaction
     /// @dev The contract derives the deposit scriptPubkey for rskAddr and takes the value of the
-    /// output paying it, through the same {PegInDerivation} helpers the registry uses at
-    /// registration. Both the derivation inputs (this proxy's address and the live powpeg script)
+    /// FIRST output paying it, through the same {PegInDerivation} helpers the registry uses at
+    /// registration. See {PegInDerivation-findFirstBtcOutputPaying} for why first-match and not the
+    /// sum, and why that is the open question here rather than in the registry.
+    /// Both the derivation inputs (this proxy's address and the live powpeg script)
     /// and the matched output come from state or from the SPV-proven bytes, so there is no
     /// argument a caller can move to change the answer.
     ///
@@ -471,10 +473,11 @@ contract PegInContract is
         view
         returns (uint256)
     {
-        bytes memory expectedPkScript = PegInDerivation.expectedDepositPkScript(
+        bytes memory expectedPkScript = PegInDerivation.depositPkScript(
             rskAddr, address(this), _bridge.getActivePowpegRedeemScript()
         );
-        (uint64 depositSats, bool found) = PegInDerivation.matchedDepositValue(btcTxSerialized, expectedPkScript);
+        (uint64 depositSats, bool found) =
+            PegInDerivation.findFirstBtcOutputPaying(btcTxSerialized, expectedPkScript);
         if (!found) {
             revert DepositOutputNotFound(rskAddr, btcTxHash);
         }
--- a/src/PegInContract.sol
+++ b/src/PegInContract.sol
@@ -335,6 +335,9 @@ contract PegInContract is
         uint256 merkleBranchPath,
         bytes32[] calldata merkleBranchHashes
     ) external payable nonReentrant whenNotHardPaused override returns (bytes32 pegInId) {
+        // Before the hash, never after: hashBtcTx would return a wtxid for the witness form, and
+        // pegInId is derived from it. See {BtcTransactionReader-WitnessSerializedTxNotAccepted}.
+        BtcTransactionReader.requireWitnessStripped(btcTxSerialized);
         bytes32 btcTxHash = BtcUtils.hashBtcTx(btcTxSerialized);
         pegInId = keccak256(abi.encodePacked(rskAddr, btcTxHash));
         if (_pegInClaims[pegInId].claimer != address(0)) {
--- a/src/PegInContract.sol
+++ b/src/PegInContract.sol
@@ -16,6 +16,7 @@ import {IPauseRegistry} from "./interfaces/IPauseRegistry.sol";
 import {IPegIn} from "./interfaces/IPegIn.sol";
 import {IPegInAddressRegistry} from "./interfaces/IPegInAddressRegistry.sol";
 import {IPegInCommitFirst} from "./interfaces/IPegInCommitFirst.sol";
+import {BtcTransactionReader} from "./libraries/BtcTransactionReader.sol";
 import {Flyover} from "./libraries/Flyover.sol";
 import {PegInDerivation} from "./libraries/PegInDerivation.sol";
 import {Quotes} from "./libraries/Quotes.sol";
@@ -455,8 +456,9 @@ contract PegInContract is
     /// @dev Two different operations, and only the first is a derivation: the deposit
     /// scriptPubkey for rskAddr is DERIVED (a formula over this proxy's address and the live
     /// powpeg script), and the amount is then READ from the FIRST output paying that script.
-    /// Both go through the {PegInDerivation} helpers the registry uses at registration. See
-    /// {PegInDerivation-findFirstBtcOutputPaying} for why first-match and not the sum, and why
+    /// Both go through the same helpers the registry uses at registration —
+    /// {PegInDerivation-depositPkScript} for the script, {BtcTransactionReader} for the lookup. See
+    /// {BtcTransactionReader-findFirstOutputPaying} for why first-match and not the sum, and why
     /// that is the open question here rather than in the registry.
     /// The derivation inputs come from state and the matched output from the SPV-proven bytes,
     /// so there is no argument a caller can move to change the answer.
@@ -478,7 +480,7 @@ contract PegInContract is
             rskAddr, address(this), _bridge.getActivePowpegRedeemScript()
         );
         (uint64 depositSats, bool found) =
-            PegInDerivation.findFirstBtcOutputPaying(btcTxSerialized, pkScript);
+            BtcTransactionReader.findFirstOutputPaying(btcTxSerialized, pkScript);
         if (!found) {
             revert DepositOutputNotFound(rskAddr, btcTxHash);
         }
--- a/src/PegInContract.sol
+++ b/src/PegInContract.sol
@@ -17,6 +17,7 @@ import {IPegIn} from "./interfaces/IPegIn.sol";
 import {IPegInAddressRegistry} from "./interfaces/IPegInAddressRegistry.sol";
 import {IPegInCommitFirst} from "./interfaces/IPegInCommitFirst.sol";
 import {Flyover} from "./libraries/Flyover.sol";
+import {PegInDerivation} from "./libraries/PegInDerivation.sol";
 import {Quotes} from "./libraries/Quotes.sol";
 import {SignatureValidator} from "./libraries/SignatureValidator.sol";
 
@@ -327,13 +328,13 @@ contract PegInContract is
     /// @inheritdoc IPegInCommitFirst
     function requestPegIn(
         address rskAddr,
-        uint256 amount,
-        bytes32 btcTxHash,
+        bytes calldata btcTxSerialized,
         bytes calldata opReturn,
         bytes32 btcBlockHash,
         uint256 merkleBranchPath,
         bytes32[] calldata merkleBranchHashes
     ) external payable nonReentrant whenNotHardPaused override returns (bytes32 pegInId) {
+        bytes32 btcTxHash = BtcUtils.hashBtcTx(btcTxSerialized);
         pegInId = keccak256(abi.encodePacked(rskAddr, btcTxHash));
         if (_pegInClaims[pegInId].claimer != address(0)) {
             revert PegInAlreadyProcessed(pegInId);
@@ -343,6 +344,7 @@ contract PegInContract is
         if (!_pegInAddressRegistry.isRegistered(rskAddr)) {
             revert AddressNotRegistered(rskAddr);
         }
+        uint256 amount = _derivePegInAmount(rskAddr, btcTxSerialized, btcTxHash);
         _requirePegInConfirmations(amount, btcTxHash, btcBlockHash, merkleBranchPath, merkleBranchHashes);
 
         uint256 fee = _configurations.calculatePegInFee(amount);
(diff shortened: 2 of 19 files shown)


src/libraries/PegInDerivation.sol
```diff
@@ -148,4 +155,58 @@ library PegInDerivation {
...
+    /// undercounting is conservative. `PegInContract.requestPegIn` uses it as the peg-in AMOUNT,
+    /// where undercounting silently drops the user's remaining outputs, so summing is the likely
+    /// correct rule there; it is not applied yet because the two call sites must move together and
+    /// because the reconciling side (`resolvePegIn`, and whether the RSK bridge sums outputs to the
+    /// derivation address) is not implemented. Revisit with settlement.
+    /// @param btcTxSerialized The witness-stripped raw deposit transaction
+    /// @param pkScript The scriptPubkey to search for, from {depositPkScript}
+    /// @return value The matched output's value in satoshis, 0 when no output matched
+    /// @return found Whether any output was locked by `pkScript`
+    function findFirstBtcOutputPaying(bytes calldata btcTxSerialized, bytes memory pkScript)
```
Luisfc68: I think there are multiple reasons to put this in the pegin contract instead of here:
- Semantically: you're not derivating anything, you're performing a lookup in a defined structure (the tx) so the function feels out of the library scope
- ~Gas efficiency: putting this in a library makes it a external call, which is more expensive, if this was a private function then the compiler will probably inline it so that version is more benefitial~
- Deployment: this is a process that is likely to change (e.g. we might be asked to allow more than one output, or a specific output, or whatever), putting this into a lib increases the cost of a potential upgrade as it forces to redeploy both the lib and the implementation

src/libraries/PegInDerivation.sol
```diff
@@ -148,4 +155,58 @@ library PegInDerivation {
...
+    /// @param rskAddr The RSK destination address the deposit address is derived for
+    /// @param pegInContract The PegInContract PROXY address mixed into step 2
+    /// @param activePowpegRedeemScript The live powpeg redeem script, read from the bridge at
+    /// call time — never stored
+    /// @return The 23-byte P2SH scriptPubkey of the deposit address
+    function depositPkScript(
+        address rskAddr,
+        address pegInContract,
+        bytes memory activePowpegRedeemScript
+    ) internal pure returns (bytes memory) {
```
Luisfc68, line 235: ~following my other comment, I'd make this external~

src/PegInContract.sol
```diff
@@ -449,6 +451,39 @@ contract PegInContract is
...
+    ///
+    /// The powpeg script is read fresh from the bridge on every call, so a federation change
+    /// rotates the expected script here exactly as it rotates the issued address. In-flight
+    /// deposits to a pre-rotation address stop matching, which is the drain-then-rotate cost
+    /// PegInDerivation documents, not a new failure mode.
+    /// @param rskAddr The RSK destination address of the peg-in
+    /// @param btcTxSerialized The witness-stripped raw deposit transaction
+    /// @param btcTxHash The txid hashed from btcTxSerialized, for the revert reason
+    /// @return The gross peg-in amount in wei
+    function _derivePegInAmount(address rskAddr, bytes calldata btcTxSerialized, bytes32 btcTxHash)
```
Luisfc68: I suggest to change the wording in the PR, this isn't a derivation, you're just parsing an extracting. Derivation suggest there is a formula or something that needs to be checked to understand it

src/PegInContract.sol
```diff
@@ -449,6 +451,39 @@ contract PegInContract is
...
+    {
+        bytes memory expectedPkScript = PegInDerivation.depositPkScript(
+            rskAddr, address(this), _bridge.getActivePowpegRedeemScript()
+        );
+        (uint64 depositSats, bool found) =
+            PegInDerivation.findFirstBtcOutputPaying(btcTxSerialized, expectedPkScript);
+        if (!found) {
+            revert DepositOutputNotFound(rskAddr, btcTxHash);
+        }
+        return uint256(depositSats) * Quotes.SAT_TO_WEI_CONVERSION;
```
Luisfc68: `SAT_TO_WEI_CONVERSION` in `Quotes` feels a bit out of place I'd put it in another lib or file with constants

test/pegin/RequestPegInTestBase.sol
Luisfc68, line 1: Besides the fixture, I'd recommend gathering a dataset of "real" transactions and test with those cases, they can be generated in regtest if needed

Luisfc68 (changes requested):

src/libraries/PegInDerivation.sol
Luisfc68: My bad, I didn't realize about the internal visibility, you can disregard that argument, I still think is worth moving this function though (the deployment argument is also invalidated assuming the rest of the lib is internal)

src/libraries/PegInDerivation.sol
Luisfc68: Similarly, you can disregard this too (asuming you don't want to centralize this in a separated contract)

test/pegin/RequestPegInTestBase.sol
Luisfc68: I can help generating the cases if needed

src/libraries/PegInDerivation.sol
Freshenext: Moved, but not into `PegInContract` — the registry calls it too (`PegInAddressRegistry._requireDepositValue`, gating registration on `MIN_DEPOSIT_SATS`), so a private function there means two copies of the same matching loop, and a drift between them means an address registers against one value and settles against another.

Your semantic point holds regardless, so it now lives in a new `src/libraries/BtcTransactionReader.sol` as `findFirstOutputPaying`. `PegInDerivation` keeps only the formula — which script a deposit must carry — and the lookup over transaction bytes sits next to it instead of inside it. Both consumers call the reader. `de4b475`.

If you would still rather see it duplicated in the two contracts, say so and I will.

src/PegInContract.sol
Freshenext: Renamed in `a0c9076`: `_derivePegInAmount` -> `_readPegInAmount`, and the doc now separates the two operations explicitly — the scriptPubkey is derived (a formula over the proxy address and the live powpeg script), the amount is read from the first output paying it.

src/PegInContract.sol
Freshenext: Agreed. `5f1a501` moves it to `Flyover.SAT_TO_WEI_CONVERSION` and `FlyoverConfigurations` re-exports it as the public value consumers should read. `Quotes` no longer owns the scale.

test/pegin/RequestPegInTestBase.sol
Freshenext: Done in `e5fe8e4` — `test/pegin/RequestPegInRealTransactions.t.sol`, nine cases against real mainnet transactions pinned by txid (`test_realTx_hashToTheirPinnedTxids` asserts the raw bytes still hash to the txid they claim, so a bad paste fails loudly rather than testing nothing).

Covered: deposit at the first output, at the second, beside a real OP_RETURN, in a single-input transaction, two outputs paying the deposit (pins the first-match rule), a flyover deposit derived for a different RSK address, and a plain P2PKH+OP_RETURN — the last three assert reverts.

`9a48240` also closes a gap the fixtures hid: `BridgeMock` discarded the hash argument, so a witness-serialized presentation of the same deposit passed in tests even though its wtxid could never be in a txid merkle tree. The mock can now confirm one specific hash.

Regtest cases would still be worth having for the SPV path, which these cannot exercise — I will take you up on that if we go after `resolvePegIn`.

src/FlyoverConfigurations.sol
```diff
@@ -49,8 +49,14 @@ contract FlyoverConfigurations is
...
-    /// fees agree with the bridge. Mirrors `Quotes.SAT_TO_WEI_CONVERSION`.
-    uint256 public constant SAT_TO_WEI_CONVERSION = 10 ** 10;
+    /// fees agree with the bridge.
+    /// @dev The public home of the scale for the commit-first path: this getter is what the SDK
+    /// and every LPS read. The value itself is defined once in {Flyover} so the peg-in contract
+    /// can share it without an external call — see that declaration for why it cannot live here.
+    /// `Quotes.SAT_TO_WEI_CONVERSION` is the legacy quote path's own copy, left alone because that
+    /// path's ABI is frozen; `test/configurations/Fee.t.sol` asserts the two agree so they cannot
+    /// drift. `PegOutContract` holds a third, private copy.
+    uint256 public constant SAT_TO_WEI_CONVERSION = Flyover.SAT_TO_WEI_CONVERSION;
```
Luisfc68: Why copying the value here? why not just using the Flyover.SAT_TO_WEI_CONVERSION directly? I'll probably be inlined anyways, also I think this could be private so the getter is not generated

Luisfc68 (approved):
I don't see blockers for the merge

P.S.: we should consider reverting requestPegIn if the serialized tx has the witness, so we don't rely on the bridge error for that scenario

src/FlyoverConfigurations.sol
Freshenext: It was not a copy: `= Flyover.SAT_TO_WEI_CONVERSION` aliases the single declaration, so the two could not disagree. But you are right that the declaration earned nothing, and its doc justified itself with a claim I did not check. It said the getter was what the SDK and every LPS read. Nothing reads it. The only caller was the drift assertion in `Fee.t.sol`, and the SDK integration test computes `10n ** 10n` itself.

Dropped in `ce0e39e` rather than made private. `_calculateFee` now reads `Flyover.SAT_TO_WEI_CONVERSION` directly, which is what you asked for, and the commit-first path has one declaration of the scale instead of a declaration plus a re-export.

The drift test moves with it: it now pins `Quotes.SAT_TO_WEI_CONVERSION` against `Flyover.SAT_TO_WEI_CONVERSION`. That is the pair that can actually diverge, since `Quotes` keeps its own copy while the legacy quote path's ABI stays frozen.

This removes `SAT_TO_WEI_CONVERSION()` from the `FlyoverConfigurations` ABI. It is not on `IFlyoverConfigurations`, so no frozen interface moves, and constants have no storage.

Freshenext:
Both of your points are addressed, @Luisfc68.

**Your P.S., the witness serialization.** Agreed, and taken: `dd81b5b`. `BtcTransactionReader.requireWitnessStripped` rejects the BIP144 form at both entry points, `PegInContract.requestPegIn` and `PegInAddressRegistry.registerAddress`, before the bytes are hashed or read.

The reliance you spotted was worse than a missing guard. `BtcUtils.getOutputs` skips the marker and flag, so it reads the same outputs from either serialization, while `hashBtcTx` double-SHA256s whatever it is handed and returns a wtxid for the witness form. One deposit could therefore present under two identities, and `pegInId` is keyed on that hash, so the already-processed check could not see the second one. Only the real bridge stopped it, because a wtxid is not in a merkle tree built from txids.

The discrimination is exact both ways. Every BIP144 encoding carries `00 01` at offsets 4 and 5. In the legacy encoding offset 4 is the input-count compactSize, never `0x00` because a transaction with no inputs is invalid, which is why BIP144 chose `0x00` as the marker. So no witness form slips through and no valid witness-stripped transaction is rejected.

`BridgeMockBlindSpot.t.sol` held the two tests that used to pass only because the bridge rejected the replay. They now assert our revert and prove the bridge is never reached, with `vm.expectCall` at count 0, verified to fail when the call does happen. Added the stripped-form control, a shorter-than-six-bytes case, a legacy negative control whose sixth byte is `0x01`, and the same pair against `registerAddress`.

Accepting the witness form later is not a matter of relaxing the check. It needs `registerBtcCoinbaseTransaction` and the witness merkle root, and the error's natspec says so, so whoever does that work revisits the guard instead of deleting it.

**`FlyoverConfigurations.SAT_TO_WEI_CONVERSION`.** Dropped in `ce0e39e`, answered in the thread.

782 tests pass. The two `DifferentialParity` failures predate this work.

PR #519, opened 2026-08-04
feat(FLY-2523): make the FlyoverConfigurations bounds editable through the timelock

Freshenext:
**Stack: 1 of 2.** Implementation only — coverage is in the stacked follow-up, [feature/FLY-2523-tests](https://github.com/rsksmart/liquidity-bridge-contract/compare/feature/FLY-2523...feature/FLY-2523-tests). Review this one first; the tests PR targets this branch, so it will retarget to `v3.0.0` automatically once this merges.

JIRA: [FLY-2523](https://rsklabs.atlassian.net/browse/FLY-2523)

## What

The configuration bounds were written once in `initialize` and never again, so widening them later needed a contract upgrade. They now move through the same queue-and-apply timelock a configuration change already uses.

- `queueBoundsChange(newMin, newMax)` and `applyBoundsChange()`, both gated on `DEFAULT_ADMIN_ROLE` and the existing delay.
- Pending bounds live in their own slot inside the existing `rsk.flyover.FlyoverConfigurations.bounds` namespace (appended, no migration), so a queued bounds change and a queued configuration change never collide.
- `BoundsChangeQueued` / `BoundsChangeApplied` mirror the visibility of the configuration events.
- Queueing again overwrites the pending pair and refreshes the eta, following the existing single-pending-change pattern.
- `getPendingBoundsChange()` exposes the queued pair, matching `getPendingChange()`.

## Decisions worth reviewing

**No plain setter.** If the admin could move a bound and a configuration value in one transaction, the bounds would stop catching admin mistakes, which is their only job — they cannot catch a mistake the same actor is free to redefine on the spot. The delay is the review window the feedback asked for.

**Bounds that would strand the active configuration outside them are rejected at apply time.** This is the open question in the ticket. The alternative — accept them and let the new bounds constrain only future changes — leaves `getPegInConfiguration()` returning values that `getPegInConfigurationBounds()` declares illegal, with no way for a reader to tell whether the pair it just read is consistent. Rejecting keeps one invariant that holds at every block and needs no qualification: *the active configuration is always within the active bounds*. The remedy for a tightening that excludes the active configuration is to move the configuration into range first — itself timelocked, and runnable during this change's own delay, so it costs no extra cycle. Widening, the case this feature exists for, is never blocked. Reasoning is in the `applyBoundsChange` NatSpec.

**Fit is checked at apply time only, not at queue time.** The active configuration can legitimately change during the delay, so a queue-time verdict would be advisory at best, and it would block the legitimate sequence of queueing a tightening now and moving the configuration into range during the wait.

**`min > max` on any field reverts** with `InvalidBounds(field, min, max)`, at queue and at apply. `initialize` now runs the same check, so no code path installs an inverted pair.

## NatSpec

The contract described the bounds as immutable deployment bounds; that language is corrected here, in `FlyoverConfigurationsRegtest`, and in the `IFlyoverConfigurations` doc comments.

⚠️ `IFlyoverConfigurations.sol` is marked frozen (S0). The edits there are **comments only** — no signature or struct is touched, so there is no ABI change — but flagging it since the file's own header calls any change to it a cross-lane event.

## Testing

`test/configurations` passes unchanged on this branch (48 tests) — this PR adds no coverage of its own by design. The new surfaces are covered in the stacked PR, which brings that suite to 76. Full suite: 773 pass, with 2 pre-existing failures in `test/differential/` (`vm.parseJsonAddress` on `.rskMainnet…`) that fail identically on `v3.0.0` and are unrelated. `solhint`: 0 errors.

### DIFF
--- a/src/FlyoverConfigurations.sol
+++ b/src/FlyoverConfigurations.sol
@@ -11,8 +11,11 @@ import {Flyover} from "./libraries/Flyover.sol";
 /// @notice The single on-chain source of the commit-first peg-in parameters (fee, confirmation
 /// tiers, amount limits) that every party reads: the SDK for its estimate, every LPS for its
 /// serve decision, and the settlement path for its amount validation. Admin changes are
-/// time-locked in two steps (queue then apply) and re-validated at both steps against immutable
-/// bounds fixed at deployment, so an admin mistake cannot set absurd values even with the role.
+/// time-locked in two steps (queue then apply) and re-validated at both steps against the active
+/// bounds, so an admin mistake cannot set absurd values even with the role. The bounds themselves
+/// are seeded at deployment and are editable by the admin only through the same two-step time
+/// lock ({queueBoundsChange} / {applyBoundsChange}), so widening or tightening them is observable
+/// for a full delay before it can take effect and never needs a contract upgrade.
 /// @dev Implements the frozen {IFlyoverConfigurations} (peg-in only); its function signatures and
 /// structs are the shared ABI every consumer depends on, so they must not be changed here.
 /// Upgradeable, ERC-7201 namespaced storage, deployed behind a TransparentUpgradeableProxy per
@@ -33,13 +36,18 @@ contract FlyoverConfigurations is
     }
 
     /// @custom:storage-location erc7201:rsk.flyover.FlyoverConfigurations.bounds
-    /// @dev Held in a separate namespace from the mutable config. Written once in `initialize`
-    /// and never again; these are the immutable deployment bounds every queued change is
-    /// re-validated against.
+    /// @dev Held in a separate namespace from the mutable config: a bounds change and a
+    /// configuration change are independent pending slots, so queueing one never clobbers the
+    /// other. `min`/`max` are seeded in `initialize` and thereafter only ever written by
+    /// `applyBoundsChange`, which the same time lock as a configuration change guards.
+    /// `timelockDelay` is not editable; it is the review window itself.
     struct FlyoverConfigurationsBounds {
         uint256 timelockDelay;
         PegConfiguration min;
         PegConfiguration max;
+        PegConfiguration pendingMin;
+        PegConfiguration pendingMax;
+        uint256 pendingEta;
     }
 
     /// @notice The version of the contract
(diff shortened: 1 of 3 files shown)


Luisfc68 (commented):
Code looks fine, but one question which was the rationale for implementing this manually and not via https://docs.openzeppelin.com/contracts/5.x/access-control#using-accessmanager or https://docs.openzeppelin.com/contracts/5.x/access-control#delayed-operation ?

Luisfc68 (approved):
Approved but pls take a look of my previous comment

Freshenext:
> Approved but pls take a look of my previous comment

I will raise a ticket to handle this.

Thanks for bringing it up!

Freshenext:
Good question. `TimelockController` is already in the repo from your own [FLY-2184](https://rsklabs.atlassian.net/browse/FLY-2184): `ChangeOwnerToTimelock.s.sol` deploys one and gives it the shared `ProxyAdmin`, so upgrades are delayed. What it deliberately does not hold is `DEFAULT_ADMIN_ROLE` on the individual contracts, so that emergency admin operations stay immediate. That boundary is why config changes cannot just reuse it: handing it the admin role would put those behind the delay too.

On top of that, [FLY-2523](https://rsklabs.atlassian.net/browse/FLY-2523) specified this shape directly — mirror the existing `queueChange`/`applyChange` on the frozen `IFlyoverConfigurations`, no plain setter, keep the review window. So the bounds flow follows the pattern the contract already had rather than introducing a second one.

`AccessManager` we do not use anywhere, and it is the thing that would actually resolve the tension: per-function delays let config be delayed while pause stays instant, with no hand-written ETA bookkeeping. Right destination, but it moves the admin surface across CollateralManagement / PegIn / PegOut / Discovery / Configurations, so it wants its own ticket. Raised as **[FLY-2613](https://rsklabs.atlassian.net/browse/FLY-2613)**, which also picks up your apply-gating point from #520.

One thing that came out of writing it up, worth knowing: pause is gated by `PAUSER_ROLE` on `PauseRegistry`, not by `DEFAULT_ADMIN_ROLE`. So the "keep pause immediate" rationale in the deploy script may not actually constrain the admin role the way it reads. [FLY-2613](https://rsklabs.atlassian.net/browse/FLY-2613) asks whoever picks it up to check that rather than inherit it.

PR #520, opened 2026-08-04
test(FLY-2523): cover the time-locked bounds change

Freshenext:
**Stack: 2 of 2.** Stacked on #519 — this PR targets `feature/FLY-2523`, not `v3.0.0`, so its diff shows tests only. GitHub retargets it to `v3.0.0` automatically once #519 merges. Review #519 first.

JIRA: [FLY-2523](https://rsklabs.atlassian.net/browse/FLY-2523)

## What

Coverage for the time-locked bounds change added in #519. Extends the two suites the ticket named, plus the getter and initialize suites for the surfaces the change added. `test/configurations` goes from 48 tests to 76.

### Timelock.t.sol

The four spec cases, mirrored for the bounds path: queue-then-apply after the delay, apply at the exact eta, apply before the delay reverts, apply with nothing queued reverts, and a second queue replacing the pending pair.

Two tests prove the change is *observable*, not merely stored: after widening, a fixed fee the old bounds rejected is accepted; after tightening, one they accepted is rejected. A third proves the pending bounds slot and the pending configuration slot are independent — queue both, apply one, and the other survives to apply cleanly.

The documented decision is pinned in both directions:
- `test_bounds_applyRejectedWhenActiveConfigFallsOutside` — a tightening that would strand the active configuration reverts with `ActiveConfigOutsideNewBounds`, and neither the bounds nor the configuration move.
- `test_bounds_tighteningSucceedsAfterMovingActiveConfigFirst` — the documented remedy works, with the configuration change running *inside* the bounds change's own delay.

Apply-time well-formedness is proven the same way the existing apply-time bounds test does it: queue a valid pair, then corrupt the stored pending min via `vm.store` (bounds namespace base + 11) so the guard is reachable.

### AdminGating.t.sol

Role gating on both entry points. One inverted-pair rejection per field (`fixedFee`, `percentageFee`, `minAmount`, `maxAmount`), each asserting the exact `InvalidBounds` payload. `min == max` is accepted as a degenerate pair that pins a field to one value. `test_noPlainBoundsSetterExists` asserts no single-step bounds setter selector is routable, so the "no code path changes bounds without the timelock" criterion is checked rather than assumed.

### Getters.t.sol / Scaffolding.t.sol

`getPendingBoundsChange()` zeroed / populated / cleared after apply. The timelock delay is unchanged by a bounds change, since the delay is the review window and is not editable. `initialize` rejects an inverted seed pair.

## Acceptance criteria

| Criterion | Test |
|---|---|
| Admin can queue and apply after the delay; both emit | `test_bounds_queueThenApplyAfterDelay_succeeds`, `test_queueBoundsChange_emitsBoundsChangeQueued`, `test_applyBoundsChange_emitsBoundsChangeApplied` |
| Applying before the delay reverts | `test_bounds_applyBeforeDelay_reverts` |
| Either function without the admin role reverts | `test_queueBoundsChange_nonAdminReverts`, `test_applyBoundsChange_nonAdminReverts` |
| `min > max` on any field reverts | four `test_queueBoundsChange_inverted*Reverts` |
| Bounds/active-config interaction documented and tested | `test_bounds_applyRejectedWhenActiveConfigFallsOutside`, `test_bounds_tighteningSucceedsAfterMovingActiveConfigFirst` |
| No path changes bounds outside the timelock | `test_noPlainBoundsSetterExists` |

## Testing

76 pass in `test/configurations`, 0 failures. Full suite: 773 pass, with 2 pre-existing failures in `test/differential/` that fail identically on `v3.0.0` and are unrelated to this change.

### DIFF
--- a/test/configurations/AdminGating.t.sol
+++ b/test/configurations/AdminGating.t.sol
@@ -215,4 +215,134 @@ contract AdminGatingTest is ConfigurationsTestBase {
         );
         config.queueChange(c);
     }
+
+    // ============================================================ bounds changes (FLY-2523)
+
+    // ------------------------------------------------------------------ role gating
+
+    function test_queueBoundsChange_nonAdminReverts() public {
+        IFlyoverConfigurations.PegConfiguration memory min = _wideMin();
+        IFlyoverConfigurations.PegConfiguration memory max = _wideMax();
+        bytes memory err = _unauthorized(stranger);
+        vm.prank(stranger);
+        vm.expectRevert(err);
+        config.queueBoundsChange(min, max);
+    }
+
+    function test_applyBoundsChange_nonAdminReverts() public {
+        _queueWideBounds();
+        vm.warp(block.timestamp + TIMELOCK_DELAY);
+        bytes memory err = _unauthorized(stranger);
+        vm.prank(stranger);
+        vm.expectRevert(err);
+        config.applyBoundsChange();
+    }
+
+    function test_admin_canQueueAndApplyBounds() public {
+        _queueWideBounds();
+        vm.warp(block.timestamp + TIMELOCK_DELAY);
+        vm.prank(owner);
+        config.applyBoundsChange();
+        (IFlyoverConfigurations.PegConfiguration memory min, ) = config
+            .getPegInConfigurationBounds();
+        assertEq(min.fixedFee, WIDE_MIN_FIXED_FEE);
+    }
+
+    /// @notice There is no way to move a bound outside the time lock: no setter exists, so the
+    /// only two selectors that write the bounds namespace are the queue/apply pair.
+    function test_noPlainBoundsSetterExists() public {
+        // A single-step setter would be the obvious name; it must not be routable.
+        (bool ok, ) = address(config).call(
+            abi.encodeWithSignature(
+                "setPegInConfigurationBounds((uint256,uint256,uint256,uint256,(uint256,uint256)[]),"
+                "(uint256,uint256,uint256,uint256,(uint256,uint256)[]))",
+                _wideMin(),
+                _wideMax()
+            )
+        );
+        assertFalse(ok, "no single-step bounds setter may exist");
+    }
+
+    // ------------------------------------------------------ inverted pair, one case per field
+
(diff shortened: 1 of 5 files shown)


test/configurations/AdminGating.t.sol
```diff
@@ -215,4 +215,134 @@ contract AdminGatingTest is ConfigurationsTestBase {
...
+    function test_queueBoundsChange_nonAdminReverts() public {
+        IFlyoverConfigurations.PegConfiguration memory min = _wideMin();
+        IFlyoverConfigurations.PegConfiguration memory max = _wideMax();
+        bytes memory err = _unauthorized(stranger);
+        vm.prank(stranger);
+        vm.expectRevert(err);
+        config.queueBoundsChange(min, max);
+    }
+
+    function test_applyBoundsChange_nonAdminReverts() public {
```
Luisfc68, line 232: maybe the apply doesn't need to be protected, think it in this way: if the queued change was already generated by the owner, then the activation after the delay doesn't have a strong reason to be restricted, operationally is simpler if only the queueing is done by the owner. IIRC that's the same approach used in OZ timelock

Luisfc68 (approved):
Same feedback as in the other PR, I think it might be better to use a battle tested approach for this feature:
https://docs.openzeppelin.com/contracts/5.x/access-control#using-accessmanager or https://docs.openzeppelin.com/contracts/5.x/access-control#delayed-operation

test/configurations/AdminGating.t.sol
Freshenext: Agreed on the reasoning, and you are right that OZ's `TimelockController` lets anyone execute once the delay has elapsed. I kept the gate because the pre-existing `applyChange` on this contract is `onlyRole(DEFAULT_ADMIN_ROLE)` as well. Ungating only the bounds path leaves one open apply and one gated apply in the same contract, which is harder to reason about than either choice consistently applied.

The safety argument is identical for both, so if we relax it I would rather do both in one change, with tests for a non-admin apply. Captured in [FLY-2613](https://rsklabs.atlassian.net/browse/FLY-2613) alongside the `AccessManager` question from #519.

PR #529, opened 2026-08-06
Feat/fly 2543

Hakob23:
## What

Add `PegOutEscrow` implementing the frozen `IPegOutEscrow` request/cancel path and getters against `FlyoverConfigurations`: fee split (no `gasFee`), dust change vs PegOut `dustThreshold`, `refundAddress(0)` → `msg.sender`, quote snapshot at request, and cancel refund while `REQUESTED`. Stub `claimPegOut` / `refundOnNoClaim` / `onSettlement`. No interface changes.

## Why

Commit-first peg-out needs an on-chain escrow so the user’s RBTC deposit is the only commitment before an LP claims. This PR lands request + cancel (and discovery getters) on top of FLY-2542 configs; claim/settlement stay stubbed for a follow-up.

## Type of Change

- [ ] Bug fix (non-breaking change)
- [x] New feature (non-breaking change)
- [ ] Breaking change
- [ ] Refactor (no intended behavior change)
- [ ] Security hardening
- [ ] Tests only
- [ ] Build/CI/Tooling
- [ ] Documentation

## Affected Areas

- [ ] PegIn flow (`registerPegIn`, `callForUser`, quote validation)
- [x] PegOut flow (`depositPegout`, `refundPegOut`, `refundUserPegOut`)
- [ ] Liquidity Provider lifecycle (register/update/status/resign/collateral/withdraw)
- [ ] Pause/operational controls
- [ ] Quote hashing/signature logic
- [ ] Bitcoin tx / PMT / merkle validation
- [ ] Deployment or upgrade scripts / Makefile targets
- [ ] CI workflows (`ci`, `fuzz`, `formal`, `slither`, `codeql`)
- [ ] Docs

## Risk & Security Review

- [x] Access control and authorization paths reviewed
- [x] Reentrancy and external call paths reviewed
- [x] Storage layout / upgrade safety reviewed (if upgradeable code changed)
- [x] Time/block/confirmation assumptions reviewed
- [ ] BTC tx output/index assumptions reviewed (if pegout validation touched)

## Test Plan

List what you ran locally and the outcome:

- [x] `npm run lint:sol` (solhint on `src/PegOutEscrow.sol`)
- [ ] `npm run lint:ts`
- [x] `npm run compile` / `forge build`
- [x] `forge test --match-path 'test/pegout-escrow/*'`
- [ ] `npm test`
- [ ] `npm run test:fuzz` (if logic/state transitions changed)
- [ ] `npm run test:invariant` (if invariant-sensitive logic changed)
- [ ] `npm run test:formal` (if formal-verification-sensitive logic changed)
- [ ] `npm run test:integration` (if integration paths changed)
- [ ] `npm run test:e2e` (if deployment/ownership flows changed)

Additional notes on test coverage, edge cases, and any tests intentionally skipped:

Unit coverage in `test/pegout-escrow/PegOutEscrow.t.sol` for request/cancel AC (happy path, zero-refund mapping, serviceability, dust refund vs fold, cancel auth/state, stubs). Claim/settlement intentionally not implemented.

## Deployment & Ops Impact

- [ ] No deployment impact
- [x] Requires deployment
- [ ] Requires upgrade
- [ ] Env/config changes required (`.env`, network RPCs, verification, etc.)
- [ ] Runbook/update instructions added to PR description

New upgradeable escrow contract; wire `initialize(admin, delay, pauseRegistry, pegOut, collateral, configurations)` after configs are live. Deploy scripts not included in this PR.

## Related Issues

- Jira: [FLY-2543](https://rsklabs.atlassian.net/browse/FLY-2543)
- Depends on: feat/FLY-2542 (peg-out FlyoverConfigurations impl)

## Reviewer Notes

- Frozen rules vs PoC: no `gasFee` in fee split; `refundAddress(0)` → `msg.sender`; quote `gasFee` / `depositConfirmations` snapshotted as `0` (not on frozen config).
- Dust read via narrow `IDustThreshold` (not on `IPegOut`); change emit is impl-only `EscrowPegOutChangePaid` (not on frozen escrow ABI).
- `claimPegOut` / `refundOnNoClaim` / `onSettlement` revert `PegOutPathNotImplemented` (TODOs marked).

## Screenshots / Logs (if applicable)

N/A

### DIFF
--- a/src/PegOutEscrow.sol
+++ b/src/PegOutEscrow.sol
@@ -14,12 +14,6 @@ import {IPegOutEscrow} from "./interfaces/IPegOutEscrow.sol";
 import {Flyover} from "./libraries/Flyover.sol";
 import {Quotes} from "./libraries/Quotes.sol";
 
-/// @dev Narrow read of PegOutContract's public dust threshold (avoid concrete import).
-/// TODO: drop this if dustThreshold is added to {IPegOut} (or another shared interface).
-interface IDustThreshold {
-    function dustThreshold() external view returns (uint256);
-}
-
 /// @title PegOutEscrow
 /// @notice Commit-first peg-out escrow. User deposits RBTC first; LPs claim and settle via
 /// PegOutContract. Fee / deadline parameters come from {IFlyoverConfigurations}.
@@ -326,16 +320,12 @@ contract PegOutEscrow is
             revert Flyover.InsufficientAmount(value, amountPlusCallFee);
         }
         changeRefund = value - amountPlusCallFee;
-        uint256 dust = _dustThreshold(address($.pegOutContract));
+        uint256 dust = $.pegOutContract.dustThreshold();
         // solhint-disable-next-line gas-strict-inequalities
         if (dust > changeRefund) {
             callFee += changeRefund;
             changeRefund = 0;
         }
-    }
-
-    function _dustThreshold(address pegOutContract) private view returns (uint256) {
-        return IDustThreshold(pegOutContract).dustThreshold();
     }
 
     function _requireRequested(PegOutEscrowStorage storage $, bytes32 requestHash) private view {
--- a/src/PegOutEscrow.sol
+++ b/src/PegOutEscrow.sol
@@ -32,12 +32,19 @@ contract PegOutEscrow is
 {
     /// @custom:storage-location erc7201:rsk.flyover.PegOutEscrow
     struct PegOutEscrowStorage {
+        /// Settlement contract (claim forwards funds here; also supplies dustThreshold).
         IPegOut pegOutContract;
+        /// Collateral registry used by claim / no-claim slash (wired now; unused until those paths).
         ICollateralManagement collateralManagement;
+        /// Active peg-out fee, bounds, deadlines, and confirmation tiers.
         IFlyoverConfigurations configurations;
+        /// Quote-shaped terms snapshotted at request (deleted on cancel / terminal states).
         mapping(bytes32 => Quotes.PegOutQuote) quotes;
+        /// Lifecycle per requestHash (`NONE` if never requested).
         mapping(bytes32 => EscrowedPegOutState) state;
-        mapping(uint256 => bytes32) idByNonce;
+        /// requestHash for each 1-based nonce; LPS rebuild after missed events.
+        mapping(uint256 => bytes32) requestHashByNonce;
+        /// How many requests have been minted; next request uses `++requestCount`.
         uint256 requestCount;
     }
 
@@ -132,7 +139,7 @@ contract PegOutEscrow is
         if (address($.configurations) == address(0)) revert FlyoverConfigurationsNotSet();
 
         IFlyoverConfigurations.PegOutConfiguration memory cfg = $.configurations.getPegOutConfiguration();
-        (uint256 amount, uint256 callFee, uint256 change) = _splitValue($, msg.value, cfg);
+        (uint256 amount, uint256 callFee, uint256 changeRefund) = _splitValue($, msg.value, cfg);
         if (amount < cfg.minAmount || amount > cfg.maxAmount) {
             revert NotServiceable(amount, cfg.minAmount, cfg.maxAmount);
         }
@@ -141,7 +148,7 @@ contract PegOutEscrow is
         requestHash = _computeRequestHash(nonce, refundAddress, destinationAddress, amount, callFee);
 
         $.state[requestHash] = EscrowedPegOutState.REQUESTED;
-        $.idByNonce[nonce] = requestHash;
+        $.requestHashByNonce[nonce] = requestHash;
 
         uint256 confirmations = $.configurations.getRequiredPegOutBtcConfirmations(amount);
         $.quotes[requestHash] =
@@ -149,9 +156,9 @@ contract PegOutEscrow is
 
         emit PegOutRequested(requestHash, refundAddress, amount, destinationAddress);
 
-        if (change > 0) {
-            emit EscrowPegOutChangePaid(requestHash, refundAddress, change);
-            _payout(refundAddress, change);
+        if (changeRefund > 0) {
+            emit EscrowPegOutChangePaid(requestHash, refundAddress, changeRefund);
+            _payout(refundAddress, changeRefund);
         }
     }
 
@@ -209,7 +216,7 @@ contract PegOutEscrow is
 
     /// @inheritdoc IPegOutEscrow
     function requestIdAt(uint256 nonce) external view override returns (bytes32) {
-        return _getStorage().idByNonce[nonce];
+        return _getStorage().requestHashByNonce[nonce];
     }
 
     // solhint-disable-next-line comprehensive-interface
@@ -297,12 +304,15 @@ contract PegOutEscrow is
         });
     }
 
-    /// @dev Frozen inverse fee split + dust-change against PegOutContract.dustThreshold.
+    /// @dev Derives principal `amount` and `callFee` from `msg.value` (frozen inverse of
+    /// `callFee = fixedFee + percentageFee·amount / 10_000`), floors `amount` to a satoshi,
+    /// then applies dust-change: residual ≥ PegOutContract.dustThreshold is returned as
+    /// `changeRefund`; smaller residual is folded into `callFee` so escrow stays fully attributed.
     function _splitValue(
         PegOutEscrowStorage storage $,
         uint256 value,
         IFlyoverConfigurations.PegOutConfiguration memory cfg
-    ) private view returns (uint256 amount, uint256 callFee, uint256 change) {
+    ) private view returns (uint256 amount, uint256 callFee, uint256 changeRefund) {
         // solhint-disable-next-line gas-strict-inequalities
         if (value <= cfg.fixedFee) {
             revert Flyover.InsufficientAmount(value, cfg.fixedFee + 1);
@@ -311,16 +321,16 @@ contract PegOutEscrow is
             / (FEE_PERCENTAGE_DENOMINATOR + cfg.percentageFee);
         amount -= amount % Quotes.SAT_TO_WEI_CONVERSION;
         callFee = $.configurations.calculatePegOutFee(amount);
-        uint256 required = amount + callFee;
-        if (value < required) {
-            revert Flyover.InsufficientAmount(value, required);
+        uint256 amountPlusCallFee = amount + callFee;
+        if (value < amountPlusCallFee) {
+            revert Flyover.InsufficientAmount(value, amountPlusCallFee);
         }
-        change = value - required;
+        changeRefund = value - amountPlusCallFee;
         uint256 dust = _dustThreshold(address($.pegOutContract));
         // solhint-disable-next-line gas-strict-inequalities
-        if (dust > change) {
-            callFee += change;
-            change = 0;
+        if (dust > changeRefund) {
+            callFee += changeRefund;
+            changeRefund = 0;
         }
     }
 
--- a/src/PegOutEscrow.sol
+++ b/src/PegOutEscrow.sol
@@ -1,0 +1,345 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.25;
+
+import {
+    AccessControlDefaultAdminRulesUpgradeable
+} from "@openzeppelin/contracts-upgradeable/access/extensions/AccessControlDefaultAdminRulesUpgradeable.sol";
+import {ReentrancyGuard} from "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
+import {EmergencyPause} from "./EmergencyPause/EmergencyPause.sol";
+import {ICollateralManagement, CollateralManagementSet} from "./interfaces/ICollateralManagement.sol";
+import {IFlyoverConfigurations} from "./interfaces/IFlyoverConfigurations.sol";
+import {IPauseRegistry} from "./interfaces/IPauseRegistry.sol";
+import {IPegOut} from "./interfaces/IPegOut.sol";
+import {IPegOutEscrow} from "./interfaces/IPegOutEscrow.sol";
+import {Flyover} from "./libraries/Flyover.sol";
+import {Quotes} from "./libraries/Quotes.sol";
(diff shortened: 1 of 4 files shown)


src/PegOutEscrow.sol
```diff
@@ -0,0 +1,345 @@
...
+    IPegOutEscrow
+{
+    /// @custom:storage-location erc7201:rsk.flyover.PegOutEscrow
+    struct PegOutEscrowStorage {
+        IPegOut pegOutContract;
+        ICollateralManagement collateralManagement;
+        IFlyoverConfigurations configurations;
+        mapping(bytes32 => Quotes.PegOutQuote) quotes;
+        mapping(bytes32 => EscrowedPegOutState) state;
+        mapping(uint256 => bytes32) idByNonce;
```
Freshenext: idByNonce - what is this

src/PegOutEscrow.sol
```diff
@@ -0,0 +1,345 @@
...
+        uint256 requestCount;
```
Freshenext, line 42: requestCount - what is this

src/PegOutEscrow.sol
```diff
@@ -0,0 +1,345 @@
...
+/// PegOutContract. Fee / deadline parameters come from {IFlyoverConfigurations}.
+/// @author Rootstock Labs
+contract PegOutEscrow is
+    AccessControlDefaultAdminRulesUpgradeable,
+    EmergencyPause,
+    ReentrancyGuard,
+    IPegOutEscrow
+{
+    /// @custom:storage-location erc7201:rsk.flyover.PegOutEscrow
+    struct PegOutEscrowStorage {
```
Freshenext, line 28: describe each field as a comment here so that we know what each are

src/PegOutEscrow.sol
```diff
@@ -0,0 +1,345 @@
...
+            depositConfirmations: 0,
+            transferConfirmations: uint16(confirmations),
+            depositAddress: destinationAddress,
+            btcRefundAddress: "",
+            lpBtcAddress: ""
+        });
+    }
+
+    /// @dev Frozen inverse fee split + dust-change against PegOutContract.dustThreshold.
+    function _splitValue(
```
Freshenext, line 314: Could we have a comment explaining what this specifically does?

Can the variables be named more detailed, for example I see a "required" -> required what?

I had to go down and see the amount - callFee < required to know about this - I think we could name it more meaningful.

Freshenext (commented):
Very formidable PR! A few comments

src/PegOutEscrow.sol
```diff
@@ -0,0 +1,345 @@
...
+import {IFlyoverConfigurations} from "./interfaces/IFlyoverConfigurations.sol";
+import {IPauseRegistry} from "./interfaces/IPauseRegistry.sol";
+import {IPegOut} from "./interfaces/IPegOut.sol";
+import {IPegOutEscrow} from "./interfaces/IPegOutEscrow.sol";
+import {Flyover} from "./libraries/Flyover.sol";
+import {Quotes} from "./libraries/Quotes.sol";
+
+/// @dev Narrow read of PegOutContract's public dust threshold (avoid concrete import).
+/// TODO: drop this if dustThreshold is added to {IPegOut} (or another shared interface).
+interface IDustThreshold {
```
Luisfc68: I think if you need this as part of the interface you could add it to the pegout contract interface, I don't see any problem

src/PegOutEscrow.sol
```diff
@@ -0,0 +1,345 @@
...
+        $.configurations = IFlyoverConfigurations(payable(configurations_));
+    }
+
+    /// @inheritdoc IPegOutEscrow
+    function requestPegOut(
+        bytes calldata destinationAddress,
+        address refundAddress
+    ) external payable override nonReentrant whenNotSoftPaused returns (bytes32 requestHash) {
+        if (destinationAddress.length == 0) revert InvalidDestination();
+        if (refundAddress == address(0)) {
```
Luisfc68, line 127: Following my comment in the interfaces PR: IMO we shouldn't allow the client to do this and make them explicitly say the refunded address

src/PegOutEscrow.sol
Luisfc68, line 314: if this is private part of the contract why does it need the storage as parameter?

Luisfc68: If the answer is to not read it twice in the same function, its ok as long as you check it is actually cheaper

src/PegOutEscrow.sol
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    function _payout(address to, uint256 amount) private {
+        (bool sent, bytes memory reason) = to.call{value: amount}("");
+        if (!sent) {
+            revert Flyover.PaymentFailed(to, amount, reason);
+        }
+    }
+
+    function _computeRequestHash(
```
Luisfc68, line 251: why this is different from a quote? the only diff between this and the quote is that the quote doens't have an LP assigned, so my question is, why the request hash can't be the quote with address(0) as LP? what is the advantage on having a separated formula?

src/PegOutEscrow.sol
```diff
@@ -0,0 +1,345 @@
...
+
+        uint256 nonce = ++$.requestCount;
+        requestHash = _computeRequestHash(nonce, refundAddress, destinationAddress, amount, callFee);
+
+        $.state[requestHash] = EscrowedPegOutState.REQUESTED;
+        $.idByNonce[nonce] = requestHash;
+
+        uint256 confirmations = $.configurations.getRequiredPegOutBtcConfirmations(amount);
+        $.quotes[requestHash] =
+            _buildQuote($, cfg, refundAddress, destinationAddress, amount, callFee, nonce, confirmations);
```
Luisfc68, line 149: you're building the quote anyways thats another reason to reuse the partial quote as request hash

src/PegOutEscrow.sol
```diff
@@ -0,0 +1,345 @@
...
+
+        emit PegOutRequested(requestHash, refundAddress, amount, destinationAddress);
+
+        if (change > 0) {
```
Luisfc68: shouldn't it be if change > dust?

src/PegOutEscrow.sol
```diff
@@ -0,0 +1,345 @@
...
+        PegOutEscrowStorage storage $,
+        bytes32 requestHash,
+        EscrowedPegOutState finalState
+    ) private {
+        $.state[requestHash] = finalState;
+        delete $.quotes[requestHash];
+    }
+
+    function _payout(address to, uint256 amount) private {
+        (bool sent, bytes memory reason) = to.call{value: amount}("");
```
Luisfc68: change this to avoid the return bomb by making a safe call
```
        address target = quote.contractAddress;
        uint256 value = quote.value;
        uint256 gasLimit = quote.gasLimit;
        bytes memory data = quote.data;
        assembly {
            success := call(gasLimit, target, value, add(data, 0x20), mload(data), 0, 0)
        }
```

src/PegOutEscrow.sol
Luisfc68: or move it to a separate file if you plan to reuse in other sources

Luisfc68 (changes requested):

src/PegOutEscrow.sol
Hakob23: The interface natspec freeze version explicitly says it, that's why I left it this way. we could stack it inside the future changes list.

src/PegOutEscrow.sol
Hakob23: Passing the storage pointer avoids a second `_getStorage()` and keeps one `$` for `configurations / dust`

src/PegOutEscrow.sol
Hakob23: At request time, no LP has claimed yet, so we don’t know who the LP will be. The stored quote still has to fill `lpRskAddress`, so we put `address(0)` there as a placeholder. That quote is only a snapshot of the terms — not the id.

The request id (requestHash) is a different, smaller hash from a fixed list of fields: chain, escrow, nonce, sender, refund address, destination, amount, fee, timestamp.

IMO we should keep them separate because the id must exist before an LP is chosen.
The freeze explicitly named it requestHash so nobody confuses it with the normal peg-out quote hash.

src/PegOutEscrow.sol
Hakob23: As discusssed here: https://github.com/rsksmart/liquidity-bridge-contract/pull/529#discussion_r3736729892

src/PegOutEscrow.sol
Hakob23: https://github.com/rsksmart/liquidity-bridge-contract/pull/529/commits/65e134469791588e8cca6234776bb06c0a9d58b2

src/PegOutEscrow.sol
Hakob23: No, this `if (changeRefund > 0)` is only the payout guard after dust is already applied in `_splitValue`.

src/PegOutEscrow.sol
Hakob23: https://github.com/rsksmart/liquidity-bridge-contract/pull/529/commits/78bf4962deb3991c3a4c16ab04d3121d4b14d6be

src/PegOutEscrow.sol
Hakob23: https://github.com/rsksmart/liquidity-bridge-contract/pull/529/commits/ee7737e7c8be90017ad4c59b6d9e9c64a77e4e79

src/PegOutEscrow.sol
Hakob23: https://github.com/rsksmart/liquidity-bridge-contract/pull/529/commits/ee7737e7c8be90017ad4c59b6d9e9c64a77e4e79

src/PegOutEscrow.sol
Hakob23: /// How many requests have been minted; next request uses `++requestCount`.

This comment has been added to the contract https://github.com/rsksmart/liquidity-bridge-contract/pull/529/commits/ee7737e7c8be90017ad4c59b6d9e9c64a77e4e79

src/PegOutEscrow.sol
Hakob23: /// requestHash for each 1-based(starting from 1 not 0) nonce; LPS rebuild after missed events

changed the name to `requestHashByNonce` for a better clarity.

https://github.com/rsksmart/liquidity-bridge-contract/pull/529/commits/ee7737e7c8be90017ad4c59b6d9e9c64a77e4e79

src/PegOutEscrow.sol
Luisfc68: Yes please

src/PegOutEscrow.sol
Luisfc68: you can still name it "requestHash" what I'm not sure is that we should use a separate formula, IMO we should maintain the quote hash, if you fix the lp address in the quote hash to address(0) you get the id that you want before the LP is chosen as you needed, doing it like this means maintaining 2 different calculations, if the quote is extended to (for example) include the address type of the destination, we'll need to remember to add it here, because the code won't break in that scenario, and if we forget we might end up in inconsistencies that can result in vulnerabilities

src/PegOutEscrow.sol
```diff
@@ -0,0 +1,355 @@
...
+    function _splitValue(
+        PegOutEscrowStorage storage $,
+        uint256 value,
+        IFlyoverConfigurations.PegOutConfiguration memory cfg
+    ) private view returns (uint256 amount, uint256 callFee, uint256 changeRefund) {
+        // solhint-disable-next-line gas-strict-inequalities
+        if (value <= cfg.fixedFee) {
+            revert Flyover.InsufficientAmount(value, cfg.fixedFee + 1);
+        }
+        amount = ((value - cfg.fixedFee) * FEE_PERCENTAGE_DENOMINATOR)
```
Luisfc68, line 323: can you explain this? you're calculating the fee here to subtract it from the amount to add it again later? I'm not sure about this, if `calculatePegOutFee` wouldn't you be forced to change this too? perhaps this should be done in configurations and not in the escrow

src/PegOutEscrow.sol
Hakob23: Follow-up kept settlement on PegOut's existing surface (`depositPegOut` / notify). See https://github.com/rsksmart/liquidity-bridge-contract/pull/549.

src/PegOutEscrow.sol
Hakob23: Zero refund address is rejected; client must set it explicitly. https://github.com/rsksmart/liquidity-bridge-contract/pull/549

src/PegOutEscrow.sol
Hakob23: Agreed — request hash is the incomplete quote hash (`lpRskAddress = 0`), not a custom preimage. https://github.com/rsksmart/liquidity-bridge-contract/pull/549

src/PegOutEscrow.sol
Hakob23: Same as above — reuse the partial quote as the request id. https://github.com/rsksmart/liquidity-bridge-contract/pull/549

src/PegOutEscrow.sol
Hakob23: Aligned with PegOut dust rule (refund when change ≥ dust). https://github.com/rsksmart/liquidity-bridge-contract/pull/549

src/PegOutEscrow.sol
Hakob23: Escrow payouts use the safe low-level call pattern. https://github.com/rsksmart/liquidity-bridge-contract/pull/549

src/PegOutEscrow.sol
Hakob23: Internal helper still takes storage for ERC-7201 access; happy to revisit style in a follow-up if you want it flattened.

src/PegOutEscrow.sol
Hakob23: Fee split now goes through shared config helpers / `calculatePegOutFee` rather than duplicating fee math. https://github.com/rsksmart/liquidity-bridge-contract/pull/549

PR #537, opened 2026-08-13
Feat/fly 2549

Hakob23:
## What

Implements short-delivery floor and escrow top-up for commit-first peg-out settlement.

- Replaces the flat `InsufficientAmount` revert on BTC delivery below `quote.value` with floor validation: `floor = amount − callFee − maxMinerFee`.
- At/above floor: settle with no top-up.
- Under floor: top up `rskRefundAddress` with `floor − deliveredAmount` and debit the LP payout by the same amount (flows conserve escrowed total).
- Snapshots admin `maxMinerFee` from `FlyoverConfigurations` at `requestPegOut` on an escrow request envelope (`quote` + `maxMinerFee`), copies it onto `PegOutRecord` at `registerClaimedPegOut`, and uses that value at `refundPegOut` so a later config change cannot move this peg-out’s floor.
- `refundPegOut` notifies escrow `FULFILLED` on successful settlement.

## Why

Ordinary BTC miner-fee variance was stranding peg-outs under the old “must deliver full `quote.value`” rule. A protocol floor plus an on-chain top-up bounds that variance. `maxMinerFee` is frozen at user request so fee/config changes cannot rewrite economics after RBTC is locked.

## Type of Change

- [ ] Bug fix (non-breaking change)
- [x] New feature (non-breaking change)
- [ ] Breaking change
- [ ] Refactor (no intended behavior change)
- [ ] Security hardening
- [ ] Tests only
- [ ] Build/CI/Tooling
- [ ] Documentation

## Affected Areas

- [ ] PegIn flow (`registerPegIn`, `callForUser`, quote validation)
- [x] PegOut flow (`depositPegout`, `refundPegOut`, `refundUserPegOut`)
- [ ] Liquidity Provider lifecycle (register/update/status/resign/collateral/withdraw)
- [ ] Pause/operational controls
- [ ] Quote hashing/signature logic
- [x] Bitcoin tx / PMT / merkle validation
- [ ] Deployment or upgrade scripts / Makefile targets
- [ ] CI workflows (`ci`, `fuzz`, `formal`, `slither`, `codeql`)
- [ ] Docs

## Risk & Security Review

- [x] Access control and authorization paths reviewed
- [x] Reentrancy and external call paths reviewed
- [x] Storage layout / upgrade safety reviewed (if upgradeable code changed)
- [x] Time/block/confirmation assumptions reviewed
- [x] BTC tx output/index assumptions reviewed (if pegout validation touched)

Notes:
- `Quotes.PegOutQuote` intentionally unchanged (cross-lane ABI / hash / signature freeze).
- Escrow ERC-7201 layout: `quotes` mapping replaced by `requests` envelope (`quote` + `maxMinerFee`); unreleased escrow path.
- `PegOutRecord` gains `maxMinerFee`; set only on claim-fed registration.
- Top-up uses the same call + balance-fallback pattern as existing payouts; must not be counted like `gasFee` in `value + callFee + gasFee` accounting.
- Under-floor no longer reverts on amount; `validatePegout` accepts under-floor txs (top-up happens only in `refundPegOut`).

## Test Plan

List what you ran locally and the outcome:

- [ ] `npm run lint:sol`
- [ ] `npm run lint:ts`
- [ ] `npm run compile`
- [ ] `npm test`
- [x] Foundry: `forge test --match-contract PegOutEscrowTest --ffi` (incl. floor/top-up cases)
- [x] Foundry: `forge test --match-contract LpRefundTest --ffi` (under-floor top-up)
- [x] Foundry: amount fuzz (`testFuzz_RefundPegOut_UnderFloor_*`, `AcceptsValidAmount`)
- [ ] `npm run test:fuzz` (if logic/state transitions changed)
- [ ] `npm run test:invariant` (if invariant-sensitive logic changed)
- [ ] `npm run test:formal` (if formal-verification-sensitive logic changed)
- [ ] `npm run test:integration` (if integration paths changed)
- [ ] `npm run test:e2e` (if deployment/ownership flows changed)

Additional notes:
- At floor / above floor (no top-up); under floor (exact top-up + LP debit + conservation); config change after request leaves snapshotted floor unchanged.
- Legacy `depositPegOut` path: `maxMinerFee` unset (0) ⇒ floor = `value − callFee`; under-floor tops up rather than reverting.

## Deployment & Ops Impact

- [ ] No deployment impact
- [x] Requires deployment
- [x] Requires upgrade
- [ ] Env/config changes required (`.env`, network RPCs, verification, etc.)
- [ ] Runbook/update instructions added to PR description

Upgradeable `PegOutEscrow` / `PegOutContract` storage layout changed (`requests` envelope; `PegOutRecord.maxMinerFee`). Treat as fresh deploy / coordinated upgrade for the unreleased commit-first stack; do not assume in-place upgrade of production quote storage without a migration plan.

## Related Issues

- Jira: [FLY-2549](https://rsklabs.atlassian.net/browse/FLY-2549)

## Reviewer Notes

- Focus: `_deliveryFloor` / top-up conservation in `refundPegOut`; request-time snapshot + claim copy; no `PegOutQuote` field for `maxMinerFee`.
- Freeze timing: request-time snapshot (not claim-time from live config) so the floor cannot move after the user locked RBTC.

### DIFF
--- a/src/PegOutContract.sol
+++ b/src/PegOutContract.sol
@@ -293,12 +293,16 @@ contract PegOutContract is
         uint256 merkleBranchPath,
         bytes32[] calldata merkleBranchHashes
     ) external nonReentrant whenNotHardPaused override {
-        Quotes.PegOutQuote memory quote = _validatePegOutTransaction(quoteHash, btcTx);
+        (Quotes.PegOutQuote memory quote, uint256 paidAmount) = _validatePegOutTransaction(quoteHash, btcTx);
         _validateBtcTxConfirmations(quote, btcTx, btcBlockHeaderHash, merkleBranchPath, merkleBranchHashes);
+
+        uint256 floor = _deliveryFloor(quote, _pegOutRegistry[quoteHash].maxMinerFee);
+        uint256 topUp = paidAmount < floor ? floor - paidAmount : 0;
 
         delete _pegOutQuotes[quoteHash];
         _pegOutRegistry[quoteHash].completed = true;
         emit PegOutRefunded(quoteHash);
+        _notifyEscrowSettlement(quoteHash, IPegOutEscrow.EscrowedPegOutState.FULFILLED);
 
         if (_shouldPenalize(quote, quoteHash, btcBlockHeaderHash)) {
             uint256 collateral = _collateralManagement.getPegOutCollateral(quote.lpRskAddress);
@@ -308,7 +312,14 @@ contract PegOutContract is
             _collateralManagement.slashPegOutCollateral(msg.sender, quote, quoteHash);
         }
 
-        uint256 refundAmount = quote.value + quote.callFee + quote.gasFee;
+        if (topUp > 0) {
+            (bool topSent,) = quote.rskRefundAddress.call{value: topUp}("");
+            if (!topSent) {
+                _increaseBalance(quote.rskRefundAddress, topUp);
+            }
+        }
+
+        uint256 refundAmount = quote.value + quote.callFee + quote.gasFee - topUp;
         (bool sent,) = quote.lpRskAddress.call{value: refundAmount}("");
         if (!sent) {
             _increaseBalance(quote.lpRskAddress, refundAmount);
@@ -364,7 +375,7 @@ contract PegOutContract is
         bytes32 quoteHash,
         bytes calldata btcTx
     ) external view override returns (Quotes.PegOutQuote memory quote) {
-        return _validatePegOutTransaction(quoteHash, btcTx);
+        (quote,) = _validatePegOutTransaction(quoteHash, btcTx);
     }
 
     /// @inheritdoc IPegOut
@@ -543,13 +554,16 @@ contract PegOutContract is
     /// Used by both validatePegout (for unbroadcasted transactions) and refundPegOut (before confirmation check).
     /// The validation expects fixed output positions in btcTx:
     /// output 0 is the payment output and output 1 is the OP_RETURN quote-hash output.
+    /// Amount shortfalls relative to the short-delivery floor are not reverted here; {refundPegOut}
+    /// tops the user up from escrow.
     /// @param quoteHash the hash of the quote being validated
     /// @param btcTx the Bitcoin transaction
     /// @return quote the PegOutQuote associated with the transaction
+    /// @return paidAmount BTC payment output value in wei
     function _validatePegOutTransaction(
         bytes32 quoteHash,
         bytes calldata btcTx
-    ) private view returns (Quotes.PegOutQuote memory quote) {
+    ) private view returns (Quotes.PegOutQuote memory quote, uint256 paidAmount) {
         if (_isQuoteCompleted(quoteHash)) revert QuoteAlreadyCompleted(quoteHash);
 
         quote = _pegOutQuotes[quoteHash];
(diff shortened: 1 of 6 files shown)


src/PegOutEscrow.sol
```diff
@@ -25,6 +25,14 @@ contract PegOutEscrow is
     ReentrancyGuard,
     IPegOutEscrow
 {
+    /// @notice Per-request escrow record: quote-shaped terms plus snapshotted floor inputs.
+    /// @dev `maxMinerFee` is frozen at {requestPegOut} from FlyoverConfigurations so a later
+    /// config change cannot move this peg-out's short-delivery floor (B8 / 10·D).
+    struct EscrowedPegOut {
+        Quotes.PegOutQuote quote;
+        uint256 maxMinerFee;
```
Luisfc68, line 33: the max miner fee is a cap to give a user a guarantee of how much they need to pay for gas, perhaps we should implement a band instead so the LP chooses the fee from there, for now is ok to have it as a single value and we can iterate it after this is merged but before releasing. In any case, my recommendation is to remove this from the `PegOutRecord` and put it in the `PegOutQuote` `gasFee` field, which is currently unused

src/PegOutContract.sol
```diff
@@ -194,6 +197,7 @@ contract PegOutContract is
         _pegOutQuotes[requestHash] = quote;
         _pegOutRegistry[requestHash].depositTimestamp = block.timestamp;
         _pegOutRegistry[requestHash].depositBlock = block.number;
+        _pegOutRegistry[requestHash].maxMinerFee = _pegOutEscrow.getMaxMinerFee(requestHash);
```
Luisfc68, line 200: Again, the idea is to leave PegOut contract's logic as similar as posible, there's no need for this contract to be aware of the max miner fee, at this point, it was already "frozen" in the quote, so this contract can keep working with the quote model

src/PegOutContract.sol
```diff
@@ -539,24 +554,40 @@ contract PegOutContract is
...
         BtcUtils.TxRawOutput[] memory outputs = BtcUtils.getOutputs(btcTx);
         _validateBtcTxNullData(outputs, quoteHash);
-        _validateBtcTxAmount(outputs, quote);
+        paidAmount = outputs[_PAY_TO_ADDRESS_OUTPUT].value * _SAT_TO_WEI_CONVERSION;
         _validateBtcTxDestination(outputs, quote);
     }
 
+    /// @notice Short-delivery floor: `amount − callFee − maxMinerFee` (walkthrough 10·D / B8).
+    /// @dev Satoshi-floors the result the same way the legacy required amount did for `value`.
+    function _deliveryFloor(
```
Luisfc68, line 580: This is incorrect. `maxMinerFee` is part of the quote, specifically in the `gasFee` field is **not** part of the amount, we shouldn't make subtractions from the amount, the miner fee was paid in the initial deposit and during the quote construction is separated from the amount itself, if its not being done in that way then the modification needs to be in the escrow not here. This change should not imply any modifications in the `PegOutContract`

src/PegOutContract.sol
```diff
@@ -600,21 +631,6 @@ contract PegOutContract is
         }
     }
 
-    /// @notice This function is used to validate the amount of the Bitcoin transaction
```
Luisfc68, line 603: This should be maintained as is, the handling of the` gasFee/minerFee` should not touch the amount validation at all.

src/PegOutContract.sol
```diff
@@ -539,24 +554,40 @@ contract PegOutContract is
...
+    /// tops the user up from escrow.
     /// @param quoteHash the hash of the quote being validated
     /// @param btcTx the Bitcoin transaction
     /// @return quote the PegOutQuote associated with the transaction
+    /// @return paidAmount BTC payment output value in wei
     function _validatePegOutTransaction(
         bytes32 quoteHash,
         bytes calldata btcTx
-    ) private view returns (Quotes.PegOutQuote memory quote) {
+    ) private view returns (Quotes.PegOutQuote memory quote, uint256 paidAmount) {
```
Luisfc68, line 566: besides all my other comments which were related to the business logic, in this case is about the code, adding multiple returns like this reduces readability, you have a function called `_validatePegOutTransaction` that for some reason now is returning a 2nd value with the paid amount, you cannot infer that by looking into the signature and you're forcing the consumer to inspect the functions body

Luisfc68: In any case as per my other comments this should remain unchanged too

Luisfc68 (changes requested):
Most of the changes in the `PegOutContract` (if not all) are not needed, in the case of the escrow is fine but include the `minerFee` in the `gasFee` field

src/PegOutEscrow.sol
Hakob23: `maxMinerFee` stays a config snapshot into `quote.gasFee` for now (band left as future work). https://github.com/rsksmart/liquidity-bridge-contract/pull/549

src/PegOutContract.sol
Hakob23: PegOut no longer carries `maxMinerFee` on its registry — fee is on the quote only. https://github.com/rsksmart/liquidity-bridge-contract/pull/549

src/PegOutContract.sol
Hakob23: Restored amount validation to `paid >= quote.value` without miner-fee subtraction from principal. https://github.com/rsksmart/liquidity-bridge-contract/pull/549

src/PegOutContract.sol
Hakob23: Same — gasFee/minerFee no longer touches amount validation. https://github.com/rsksmart/liquidity-bridge-contract/pull/549

src/PegOutContract.sol
Hakob23: Validation flow kept on the existing `_validatePegOutTransaction` path in the follow-up. https://github.com/rsksmart/liquidity-bridge-contract/pull/549

Hakob23:
Fixes are applied in https://github.com/rsksmart/liquidity-bridge-contract/pull/549

